PR #3, opened 2019-10-04
Changes to support deployment

jpslav:
Changes to support deployment

These changes enable production environment creation with the deployment scripts.


lib/tasks/install_secrets.rake
```diff
@@ -0,0 +1,84 @@
+require 'aws-sdk-ssm'
```
jpslav, line 1: largely copied from open-search

steveburkett:
@jpslav Can we get a readme write-up on how all this works, production and locally?

lib/tasks/install_secrets.rake
steveburkett: seems like it might be a good idea to pull the code out of the rake task into a generalized and shared spot, with unit tests?  e.g., in the aws ruby gem?

steveburkett (approved):

PR #6, opened 2019-10-20
allow 'highlights' path prefixes for use with CloudFront

jpslav:
allow 'highlights' path prefixes for use with CloudFront

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -5,7 +5,9 @@ require "rails"
 # This is the railtie middleware stack.
 # !! Have removed some of these to improve API performance. !!
 require "active_model/railtie"
-# require "active_job/railtie"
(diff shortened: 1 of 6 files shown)


config/application.rb
```diff
@@ -5,7 +5,9 @@
 # This is the railtie middleware stack.
 # !! Have removed some of these to improve API performance. !!
 require "active_model/railtie"
-# require "active_job/railtie"
+# can comment out active_job/railtie after next 5.2.x release, see
+# https://github.com/rails/rails/pull/35896#issuecomment-535569136
```
steveburkett, line 9: dang, kinda lame we have to pull in active_job baggage...but sounds good, great find!

steveburkett (approved):

PR #8, opened 2019-10-22
initial migrations for highlights

steveburkett:
Highlight table schema and basic tests

Highlighting doc https://docs.google.com/document/d/1eUzJ6YDwK25K8gHllXljr5ELKE06tWjAh6hXIAYcqsw/edit#heading=h.v01zz2q3y1e7

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/602

### DIFF
--- a/app/models/highlight.rb
+++ b/app/models/highlight.rb
@@ -8,6 +8,7 @@ class Highlight < ApplicationRecord
   validates_format_of :color, with: /\A#?(?:[a-f0-9]{3}){1,2}\z/
   validates_presence_of :user_uuid,
                         :source_type,
+                        :source_id,
                         :anchor,
                         :highlighted_content,
                         :location_strategies
--- a/app/models/highlight.rb
+++ b/app/models/highlight.rb
@@ -1,0 +1,11 @@
+# OpenStax Highlights
+#
+# This is the main model for the Highlights API. This model represents the
+# highlight or note used by the API
+#
+# See https://docs.google.com/document/d/1eUzJ6YDwK25K8gHllXljr5ELKE06tWjAh6hXIAYcqsw/edit#heading=h.v01zz2q3y1e7 for more information.
+class Highlight < ApplicationRecord
+  validates_format_of :color, with: /\A#?(?:[A-F0-9]{3}){1,2}\z/i
+
+  enum source_type: [:openstax_page]
+end
--- a/app/models/highlight.rb
+++ b/app/models/highlight.rb
@@ -5,7 +5,25 @@
 #
 # See https://docs.google.com/document/d/1eUzJ6YDwK25K8gHllXljr5ELKE06tWjAh6hXIAYcqsw/edit#heading=h.v01zz2q3y1e7 for more information.
 class Highlight < ApplicationRecord
-  validates_format_of :color, with: /\A#?(?:[A-F0-9]{3}){1,2}\z/i
+  validates_format_of :color, with: /\A#?(?:[a-f0-9]{3}){1,2}\z/
+  validates_presence_of :user_uuid,
+                        :source_type,
+                        :anchor,
+                        :highlighted_content,
+                        :location_strategies
 
   enum source_type: [:openstax_page]
+
+  before_validation :normalize_color, :normalize_source_ids
+
+  def normalize_color
+    self.color = color&.downcase
+  end
+
+  def normalize_source_ids
+    if openstax_page?
+      self.source_id = source_id&.downcase
+      source_parent_ids.map!(&:downcase)
+    end
+  end
 end
--- a/db/migrate/20191017173513_create_highlights.rb
+++ b/db/migrate/20191017173513_create_highlights.rb
@@ -1,0 +1,24 @@
+# OpenStax Highlights
+class CreateHighlights < ActiveRecord::Migration[5.2]
+  def change
+    create_table :highlights do |t|
+      t.uuid :user_uuid, null: false, index: true
+      t.integer :source_type, null: false, default: 0
+      t.uuid :source_uuid, null: false
+      t.jsonb :source_metadata, null: true
+      t.text :source_parent_ids, array: true, default: []
+      t.text :anchor, null: false
+      t.text :highlighted_content, null: false
+      t.text :annotation, null: true
+      t.string :color, null: false
+      t.text :source_pagingation_order, null: true
+      t.integer :order_in_source, null: true
+      t.jsonb :location_strategies, null: false
+
+      t.timestamps null: false
+    end
+  end
+end
+
+
+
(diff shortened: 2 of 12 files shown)


app/models/highlight.rb
```diff
@@ -0,0 +1,11 @@
+# OpenStax Highlights
+#
+# This is the main model for the Highlights API. This model represents the
+# highlight or note used by the API
+#
+# See https://docs.google.com/document/d/1eUzJ6YDwK25K8gHllXljr5ELKE06tWjAh6hXIAYcqsw/edit#heading=h.v01zz2q3y1e7 for more information.
+class Highlight < ApplicationRecord
+  validates_format_of :color, with: /\A#?(?:[A-F0-9]{3}){1,2}\z/i
```
jpslav: can we normalize the color to be lowercase and remove the case insensitive match? e.g. with a `before_validation`

db/migrate/20191017173513_create_highlights.rb
```diff
@@ -0,0 +1,24 @@
+# OpenStax Highlights
+class CreateHighlights < ActiveRecord::Migration[5.2]
+  def change
+    create_table :highlights do |t|
```
jpslav: Make the `id` be a UUID

```ruby
create_table :highlights, id: :uuid do |t|
```

Can also set generators to do this by default in `application.rb`:

```ruby
config.generators do |g|
  g.orm :active_record, primary_key_type: :uuid
end
```

db/migrate/20191017173513_create_highlights.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.uuid :user_uuid, null: false, index: true
+      t.integer :source_type, null: false, default: 0
+      t.uuid :source_uuid, null: false
```
jpslav: not sure if all sources will be identified by a UUID -- pages will but could easily not have been if we wanted `uuid@version` as the source ID.  So think this should be `source_id` and be string.  Also gels better with `source_parent_ids`.

db/migrate/20191017173513_create_highlights.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.jsonb :source_metadata, null: true
+      t.text :source_parent_ids, array: true, default: []
+      t.text :anchor, null: false
+      t.text :highlighted_content, null: false
+      t.text :annotation, null: true
+      t.string :color, null: false
+      t.text :source_pagingation_order, null: true
```
jpslav: Can we take out `_pagination_`?

db/migrate/20191017173513_create_highlights.rb
```diff
@@ -0,0 +1,24 @@
+# OpenStax Highlights
+class CreateHighlights < ActiveRecord::Migration[5.2]
+  def change
+    create_table :highlights do |t|
+      t.uuid :user_uuid, null: false, index: true
+      t.integer :source_type, null: false, default: 0
```
jpslav: index it?

db/migrate/20191017173513_create_highlights.rb
jpslav: Can we have `source_type`-specific normalization of `source_id` and `source_parent_ids`, i.e. to downcase them for openstax_page type?

db/migrate/20191017173513_create_highlights.rb
```diff
@@ -0,0 +1,24 @@
...
+      t.uuid :source_uuid, null: false
+      t.jsonb :source_metadata, null: true
+      t.text :source_parent_ids, array: true, default: []
```
jpslav, line 13: add index (`gin`) like in https://coderwall.com/p/sud9ja/rails-4-the-postgresql-array-data-type ?

jpslav (changes requested):

app/models/highlight.rb
```diff
@@ -0,0 +1,29 @@
...
+# highlight or note used by the API
+#
+# See https://docs.google.com/document/d/1eUzJ6YDwK25K8gHllXljr5ELKE06tWjAh6hXIAYcqsw/edit#heading=h.v01zz2q3y1e7 for more information.
+class Highlight < ApplicationRecord
+  validates_format_of :color, with: /\A#?(?:[a-f0-9]{3}){1,2}\z/
+  validates_presence_of :user_uuid,
+                        :source_type,
+                        :anchor,
+                        :highlighted_content,
+                        :location_strategies
```
jpslav, line 14: `source_id` too?

db/migrate/20191017173513_create_highlights.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.text :annotation, null: true
+      t.string :color, null: false
+      t.text :source_order, null: true
+      t.integer :order_in_source, null: true
+      t.jsonb :location_strategies, null: false
+
+      t.timestamps null: false
+    end
+
+    add_index :highlights, :source_parent_ids, using: 'gin'
```
jpslav, line 29: Do we need multi-column indexes?  For when we query where user UUID == blah AND source_parent_IDs has blah?

spec/models/highlight_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    describe 'before validation hooks' do
+      it 'color will be downcased' do
+        record = described_class.new(color: '#C0C0C0')
+        record.validate
+        expect(record.color).to eq '#c0c0c0'
+      end
+
+      it 'source_parent_ids will be downcased' do
```
jpslav, line 54: Put these inside a `context 'openstax_page source_type'` block

spec/models/highlight_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        record = described_class.new(source_type: 0, source_parent_ids: ['ABC'])
```
jpslav: Use `described_class.openstax_page.new` to be more readable than `source_type: 0`?

jpslav (changes requested):

spec/models/highlight_spec.rb
steveburkett: yes indeed.  that's a cool rails trick!

db/migrate/20191017173513_create_highlights.rb
steveburkett: yea, we’ll def need those..but i figured we’d wait on that to find out what queries we’d make.  single indexes help a lot (a postgres query would use the first field value’s index and then brute force after that..but that’s usually pretty good!).   optimal is a compound index on exactly the where fields…but i dont know what those are right now…which is why i had just primary fields.  if you know what compound field we are, in the future, going to query on..that’s great if you want to add these now

jpslav (approved):

PR #10, opened 2019-10-29
add highlight

steveburkett:
Adding a highlight and swagger docs

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/606

image


Gemfile
```diff
@@ -45,6 +45,8 @@ gem 'dotenv-rails'
 
 gem "openstax_swagger", github: 'openstax/swagger-rails', ref: '0ad77d306d2'
 
+gem 'active_model_serializers'
```
jpslav: what is this used for?  guessing unused `HighlightSerializer` class?

app/controllers/api/v0/base_controller.rb
```diff
@@ -1,4 +1,11 @@
 class Api::V0::BaseController < ApplicationController
   include Swagger::Blocks
   include OpenStax::Swagger::Bind
+
+  include Response
+  include ExceptionHandler
+
+  def binding_error(status_code:, messages:)
```
jpslav, line 19: make `protected`?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -0,0 +1,58 @@
+class Api::V0::HighlightsController < Api::V0::BaseController
+  # before_action :set_highlight, only: [:show, :update, :destroy]
+
+  swagger_path '/highlights' do
+    operation :post do
+      key :summary, 'Add a highlight'
+      key :description, 'Add a OpenStax highlight or note.'
```
jpslav: Don't need to say an "OpenStax highlight".  And "or note" isn't quite right either b/c you can't add a note without a highlight.  If you want something longer than "Add a highlight", can do "Add a highlight with an optional note" or something.

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      key :operationId, 'addHighlight'
+      key :produces, [
+        'application/json'
+      ]
+      key :tags, [
+        'Highlights'
+      ]
+      # security do
+      #   key :api_id, []
+      # end
```
jpslav, line 17: later we'll need to figure out how to describe the cookie auth

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+      end
+      response 200 do
+        key :description, 'Success.  Returns the created highlights.'
+        schema do
+          key :'$ref', :Highlight
+        end
+      end
+      extend Api::V0::SwaggerResponses::AuthenticationErrorId
+      extend Api::V0::SwaggerResponses::ForbiddenErrorId
```
jpslav: Can this return forbidden?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      extend Api::V0::SwaggerResponses::ServerError
+    end
+  end
+
+  def create
+    binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
+    render(json: error, status: error.status_code) and return if error
+
+    @highlight = Highlight.create!(highlight_params)
```
jpslav: The schema validation above can error but so can this right? (if validations fail).  Do we need to return those errors as well? As written would this 500 if there's a validation error? (I think it would)

API schema errors can be different from model errors, generally.

Also, this is where we'd set the uuid of the user from the `current_user` (or similar object based on the SSO cookie)

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    render json: @highlight, status: :created
+  end
+
+  private
+
+  def highlight_params
+    params.require(:highlight).permit(:user_uuid, :source_type, :source_id, :source_metadata,
+                                      :source_parent_ids, :anchor, :highlighted_content, :annotation,
+                                      :color, :source_order, :order_in_source, :location_strategies => [])
```
jpslav: I think these lines in `bind` permit the fields defined in the swagger model, so we shouldn't have to do this extra `permit` call here? https://github.com/openstax/swagger-rails/blob/27bb0d519026707b249fde2fb101f981e508389c/lib/openstax/swagger/bind.rb#L13-L15

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      extend Api::V0::SwaggerResponses::ServerError
+    end
+  end
+
+  def create
+    binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
+    render(json: error, status: error.status_code) and return if error
+
+    @highlight = Highlight.create!(highlight_params)
+    render json: @highlight, status: :created
```
jpslav: Documentation says we return 200 (ok), but this says 201 (created).

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,51 @@
+module Api::V0::Swagger::Models::Highlight
+  include Swagger::Blocks
+  include SwaggerBlocksExtensions
+
+  swagger_schema :Highlight do
+    key :required, [:id, :user_uuid]
+  end
+
+  swagger_schema :NewHighlight do
+    key :required, [:user_uuid, :source_type, :source_id, :anchor, :highlighted_content, :color, :location_strategies]
```
jpslav: I think the `user_uuid` shouldn't be required in the posted body for a new highlight.  Instead we should just get it from the authentication data (from `current_user` which would get it from the SSO cookie).  We can't trust a `user_uuid` set in the posted body.

Which also means it shouldn't be required up on line 6.

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    property :source_type do
+      key :type, :string
+      key :enum, ['openstax_page']
+    end
+    property :source_id do
+      key :type, :string
+      key :format, 'uuid'
```
jpslav: not necessarily a uuid for all `source_type`s

app/models/highlight.rb
```diff
@@ -5,7 +5,9 @@
 #
 # See https://docs.google.com/document/d/1eUzJ6YDwK25K8gHllXljr5ELKE06tWjAh6hXIAYcqsw/edit#heading=h.v01zz2q3y1e7 for more information.
 class Highlight < ApplicationRecord
-  validates_format_of :color, with: /\A#?(?:[a-f0-9]{3}){1,2}\z/
+  VALID_COLOR = '\A#?(?:[a-f0-9]{3}){1,2}\z'
```
jpslav: see slack thread

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,51 @@
...
+      key :type, :string
+    end
+    property :color do
+      key :type, :string
+      key :pattern, '#?(?:[a-f0-9]{3}){1,2}'
+    end
+    property :location_strategies do
+      key :type, :array
+      items do
+        key :type, :string
```
jpslav: `:object` I think?

app/controllers/api/v0/swagger_blocks_extensions.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  module ClassMethods
+    def add_properties(*schema_names, &block)
+      schema_names.each do |schema_name|
+        swagger_schema schema_name do
+          instance_eval(&block)
+        end
+      end
+    end
+  end
```
jpslav: This stuff is already in swagger-rails: https://github.com/openstax/swagger-rails/blob/27bb0d519026707b249fde2fb101f981e508389c/lib/openstax/swagger/swagger_blocks_extensions.rb

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+    end
+  end
+
+  add_properties(:NewHighlight, :Highlight) do
+    property :id do
+      key :type, :string
+      key :format, 'uuid'
+      key :readOnly, true
```
jpslav: not read only in `NewHighlight` as it can be set by the client

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,51 @@
...
+      key :description, 'Internally set UUID'
```
jpslav: Doesn't have to be internally set

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+    property :user_uuid do
+      key :type, :string
+      key :format, 'uuid'
+      key :description, 'The ID of the user accessing the highlight'
+    end
```
jpslav: like above, `user_uuid` not set on add highlight

app/controllers/api/v0/swagger_responses.rb
```diff
@@ -0,0 +1,63 @@
+module Api::V0::SwaggerResponses
+  module AuthenticationErrorId
+    def self.extended(base)
+      base.response 401 do
+        key :description, 'Not authorized.  Happens when API ID or origin are not set.'
```
jpslav: These descriptions are not right for this app (API ID, origin, API token are copy paste and n/a)

app/controllers/concerns/response.rb
```diff
@@ -0,0 +1,5 @@
+module Response
+  def json_response(object, status = :ok)
+    render json: object, status: status
+  end
```
jpslav: What does the `json_response` wrapper gain us?

app/serializers/highlight_serializer.rb
```diff
@@ -0,0 +1,3 @@
+class HighlightSerializer < ActiveModel::Serializer
+  attributes :id, :source_type, :user_uuid, :source_id, :location_strategies
+end
```
jpslav: Unused?

config/initializers/highlight_override.rb
```diff
@@ -0,0 +1,31 @@
+# Monkey patch the generated Bindings for the swagger Highlight model
+require_relative '../../app/bindings/api/v0/bindings/new_highlight'
```
jpslav: This require isn't needed because the bindings are autoloaded

config/initializers/highlight_override.rb
```diff
@@ -0,0 +1,31 @@
...
+      location_strategies = to_hash[:location_strategies]
+      location_strategies.all? do |location_strategy|
+        strategy = JSON.parse(location_strategy)
+        case strategy['type']
+        when 'TextPositionSelector'
+          keys_matcher(%w(start end), strategy.keys)
+        when 'XpathRangeSelector'
+          keys_matcher(%w(endContainer endOffset startContainer startOffset),
+                       strategy.keys)
+        else
```
jpslav: Instead of validating by checking for keys, was thinking we'd define schemas for `TextPositionSelector` and `XpathRangeSelector` and generate bindings for them.  This way we don't have to hard code properties in the monkey patch and we get a more full validity check and error reporting.

Then overriding `location_strategies=`:

```ruby
def location_strategies=(array)
  @location_strategies = array.map do |item|
    case item[:type]
    when 'TextPositionSelector'
      Api::V0::Bindings::TextPositionSelector.new(item)
    when 'XpathRangeSelector'
      Api::V0::Bindings::XpathRangeSelector.new(item)
    end
  end
end
```

Then `valid?` can delegate to the `valid?` method on the strategy bindings:

```ruby
def valid?
  old_valid? && !location_strategies.empty? && location_strategies.all?(&:valid?)
end
```

Additionally `list_invalid_properties` (used in bind to report errors) can be extended:

```ruby
alias_method :old_list_invalid_properties, :list_invalid_properties
def list_invalid_properties
  invalid_properties = old_list_invalid_properties

  if location_strategies.empty?
    invalid_properties.push('invalid value for "location_strategies", location_strategies cannot be empty.')
  end

  location_strategies.each do |strategy|
    strategy.list_invalid_properties.each do |strategy_invalid_property|
      invalid_properties.push("invalid value for location strategy #{strategy.type}: #{strategy_invalid_property}")
    end
  end 

  invalid_properties
end
```

jpslav (changes requested):

Gemfile
steveburkett: @jpslav this is used to serialize the json response in #create

app/serializers/highlight_serializer.rb
steveburkett: this is used to serialize the highlight in the #create response body

app/controllers/concerns/response.rb
steveburkett: not a lot. i had that in there with a few other things, but pared it down...it can be removed

config/initializers/highlight_override.rb
steveburkett: using class_eval to autoload the NewHighlight binding prior to monkey patching

app/serializers/highlight_serializer.rb
steveburkett: removed in leu of using the binding

config/initializers/highlight_override.rb
steveburkett: this was changed to a class_eval, so will remove the require_relative

app/controllers/api/v0/highlights_controller.rb
steveburkett: The way i had this, it would return a message and unprocessable_entity

```
    rescue_from ActiveRecord::RecordInvalid, ActionController::ParameterMissing do |e|
      render json: { message: e.message }, status: :unprocessable_entity
    end
```

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  def create
+    binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
+    render(json: error, status: error.status_code) and return if error
+
+    # temporary test user_uuid.  will be removed when auth is set up
+    temp_user_uuid = '55783d49-7562-4576-a626-3b877557a21f'
+    created_highlight = Highlight.create!(binding.to_hash.merge(user_uuid: temp_user_uuid))
+
+    response_binding = Api::V0::Bindings::Highlight.create_from_model(created_highlight)
+    render json: response_binding.to_hash, status: :created
```
jpslav: Do you need to say `.to_hash`?  Thought `render json: blah` would try to call `to_json` on `blah`?

app/controllers/api/v0/swagger/docs_controller.rb
```diff
@@ -38,6 +38,9 @@ class Api::V0::Swagger::DocsController < ApplicationController
   end
 
   SWAGGERED_CLASSES = [
+    Api::V0::HighlightsController,
+    Api::V0::Swagger::Models::Highlight,
+    Api::V0::Swagger::Models::Error,
```
jpslav, line 43: Add the other new models for the location strategies (unless it picks them up from inside `Models::Highlight`)

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,75 @@
...
+      key :enum, ['openstax_page']
+    end
+    property :source_id do
+      key :type, :string
+    end
+    property :anchor do
+      key :type, :string
+    end
+    property :highlighted_content do
+      key :type, :string
```
jpslav, line 73: Doesn’t have to be right now, but before we declare highlighting victory, please add descriptions to all the fields, e.g. https://github.com/a15k/interactions-api/blob/master/app/controllers/api/v0/swagger_models/flag.rb#L28

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,75 @@
...
+    property :color do
+      key :type, :string
+      # remove the anchors because swagger-codegen always escapes them
+      key :pattern, ::Highlight::VALID_COLOR.inspect[1..-2]
+    end
+    property :location_strategies do
+      key :type, :array
+      items do
+        key :type, :object
+      end
```
jpslav, line 89: In the description for `location_strategies`, please reference that the items should have a schema matching the strategy schemas that have been defined.

config/initializers/highlight_override.rb
```diff
@@ -0,0 +1,54 @@
+# Monkey patch the generated Bindings for the swagger Highlight model
```
jpslav, line 1: When we make v1 we’ll have to put the v1 override somewhere.  If our scripts don’t blow away the bindings directory, would you want to put this file directly in `/app/bindings/api/v0/bindings`?

config/initializers/highlight_override.rb
```diff
@@ -0,0 +1,54 @@
...
+
+Api::V0::Bindings::NewHighlight.class_exec do
+  def valid_location_strategies?
+    location_strategies.present? &&
```
jpslav, line 5: The binding already checks against `location_strategies.nil?`

config/initializers/highlight_override.rb
```diff
@@ -0,0 +1,54 @@
...
+      location_strategies.none? { |item| item.nil? } &&
```
jpslav: `location_strategies.none?(&:nil?)` if you want to match the succinctness of the next line

config/initializers/highlight_override.rb
```diff
@@ -0,0 +1,54 @@
...
+  alias_method :old_list_invalid_properties, :list_invalid_properties
+  def list_invalid_properties
+    invalid_properties = old_list_invalid_properties
+
+    if location_strategies.blank?
+      invalid_properties.push('invalid value for "location_strategies", location_strategies cannot be empty.')
+    end
+
+    if location_strategies.any? { |item| item.nil? }
+      invalid_properties.push('invalid strategy detected')
```
jpslav, line 36: `location_strategies.any?(&:nil?)` and then maybe more meaning in the error message, e.g. ‘location strategies cannot be null’?  and generally maybe should make these error messages look a little more similar (this one is different than the above’s format, for example)

config/initializers/highlight_override.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  alias_method :old_list_invalid_properties, :list_invalid_properties
+  def list_invalid_properties
+    invalid_properties = old_list_invalid_properties
+
+    if location_strategies.blank?
+      invalid_properties.push('invalid value for "location_strategies", location_strategies cannot be empty.')
+    end
+
+    if location_strategies.any? { |item| item.nil? }
```
jpslav: this one will blow up if the array is nil.  you catch this below with the `&.` — maybe would be good to move this check into that next loop since we also check there for `strategy.nil?`?

jpslav (changes requested):

app/controllers/api/v0/swagger/docs_controller.rb
jpslav: nvm should pick them up

app/controllers/api/v0/highlights_controller.rb
steveburkett: correct.  the to_hash isnt needed

config/initializers/highlight_override.rb
steveburkett: present checks for both nil? and empty array (which is relevant here...if the location strategies are []

steveburkett: present checks for both nil? and empty array (which is relevant here...if the location strategies are []

steveburkett: present checks for both nil? and empty array (which is relevant here...if the location strategies are [])

config/initializers/highlight_override.rb
steveburkett: good call.   i think this is a good discussion point sometime.  both are idiomatic ruby, but one is more terse and potentially misunderstood by some dev?  it was just a little bit longer and more self evident which is why i used this form.  but both are fantastic.  will change to as you suggested to match the next line, which makes good sense.

steveburkett: good call.   i think this is a good discussion point sometime.  both are idiomatic ruby, but one is more terse and potentially misunderstood by some dev?  it was just a little bit longer and more self evident which is why i used this form.  but both are fantastic.  will change to as you suggested to match the next line, which makes good sense.

steveburkett: good call.   i think this is a good discussion point sometime.  both are idiomatic ruby, but one is more terse and potentially misunderstood by some dev?  it was just a little bit longer and more self evident which is why i used this form.  but both are fantastic.  will change to as you suggested to match the next line, which makes good sense.

steveburkett: good call.   i think this is a good discussion point sometime.  both are idiomatic ruby, but one is more terse and potentially misunderstood by some dev?  it was just a little bit longer and more self evident which is why i used this form.  but both are fantastic.  will change to as you suggested to match the next line, which makes good sense.

steveburkett: good call.   i think this is a good discussion point sometime.  both are idiomatic ruby, but one is more terse and potentially misunderstood by some dev?  it was just a little bit longer and more self evident which is why i used this form.  but both are fantastic.  will change to as you suggested to match the next line, which makes good sense.

config/initializers/highlight_override.rb
steveburkett: in this case, which is covered by a test, is when a location strategy is passed in that's not one of the two we recognize.  it then gets set to nil (in the setter) which is then handled here (perhaps an empty array element was passed in, or as described here, an invalid foobar one.  

in this case, it returns "invalid strategy detected" without making the code unnecessarily complex.  

that was the idea.

config/initializers/highlight_override.rb
steveburkett: let's chat about this?

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,92 @@
...
+      key :type, :string
+      key :description, 'The start to the text position selector.'
+    end
+    property :end do
+      key :type, :string
+      key :description, 'The end to the text position selector.'
+    end
+    property :type do
+      key :type, :string
+      key :description, 'The type for the text position selector.'
```
jpslav, line 17: I think this and the other strategy type should be an enum that requires them to be the defined string (i.e. this field is only allowed to have one value "TextPositionSelector")

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -0,0 +1,92 @@
...
+      key :type, :string
+      key :description, 'The start container for the xpath range selector.'
+    end
+    property :startOffset do
+      key :type, :integer
+      key :description, 'The start offset for the xpath range selector.'
+    end
+    property :type do
+      key :type, :string
+      key :description, 'The type for the xpath range selector.'
```
jpslav, line 41: ditto enum comment above

jpslav (approved):

PR #11, opened 2019-11-05
get user uuid from sso cookie

jpslav:
* get user uuid from sso cookie
* provide an rspec stub method for setting the current user's UUID
* provide a way to stub the current user's UUID in development by using a `STUBBED_USER_UUID` environment variable.

### DIFF
--- a/app/controllers/api/v0/highlights_controller.rb
+++ b/app/controllers/api/v0/highlights_controller.rb
@@ -1,4 +1,6 @@
 class Api::V0::HighlightsController < Api::V0::BaseController
+  before_action :render_unauthorized_if_no_current_user
+
   # before_action :set_highlight, only: [:show, :update, :destroy]
 
   swagger_path '/highlights' do
@@ -37,14 +39,12 @@ class Api::V0::HighlightsController < Api::V0::BaseController
   end
 
   def create
-    binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
+    inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
     render(json: error, status: error.status_code) and return if error
 
-    # temporary test user_uuid.  will be removed when auth is set up
-    temp_user_uuid = '55783d49-7562-4576-a626-3b877557a21f'
-    created_highlight = Highlight.create!(binding.to_hash.merge(user_uuid: temp_user_uuid))
+    model = inbound_binding.create_model!(user_uuid: current_user_uuid)
 
-    response_binding = Api::V0::Bindings::Highlight.create_from_model(created_highlight)
+    response_binding = Api::V0::Bindings::Highlight.create_from_model(model)
     render json: response_binding, status: :created
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,13 +1,21 @@
 class ApplicationController < ActionController::API
 
+  protected
+
   include RescueFromUnlessLocal
 
   def current_user_uuid
-    if Rails.environment.development? && ENV['FORCED_USER_UUID']
-      ENV['FORCED_USER_UUID']
-    else
-      OpenStax::Accounts::Sso.user_uuid(request)
+    @current_user_uuid ||= begin
+      if Rails.env.development? && ENV['STUBBED_USER_UUID']
+        ENV['STUBBED_USER_UUID']
+      else
+        OpenStax::Accounts::Sso.user_uuid(request)
+      end
     end
+  end
+
+  def render_unauthorized_if_no_current_user
+    head :unauthorized if current_user_uuid.nil?
   end
 
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,4 +2,12 @@ class ApplicationController < ActionController::API
 
   include RescueFromUnlessLocal
 
+  def current_user_uuid
+    if Rails.environment.development? && ENV['FORCED_USER_UUID']
+      ENV['FORCED_USER_UUID']
+    else
+      OpenStax::Accounts::Sso.user_uuid(request)
+    end
+  end
+
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,6 +9,11 @@ class ApplicationController < ActionController::API
       if Rails.env.development? && ENV['STUBBED_USER_UUID']
         ENV['STUBBED_USER_UUID']
       else
+        if ENV['STUBBED_USER_UUID']
+          Rails.logger.warn("`STUBBED_USER_UUID` environment variable is set but not used in " \
+                            "the #{Rails.env} environment.")
+        end
+
         OpenStax::Accounts::Sso.user_uuid(request)
       end
     end
--- a/config/initializers/highlight_override.rb
+++ b/config/initializers/highlight_override.rb
@@ -53,3 +53,9 @@ Api::V0::Bindings::Highlight.class_exec do
     new(model.attributes)
   end
 end
+
+Api::V0::Bindings::NewHighlight.class_exec do
+  def create_model!(user_uuid:)
+    Highlight.create!(to_hash.merge(user_uuid: user_uuid))
+  end
+end
--- a/spec/support/user_helpers.rb
+++ b/spec/support/user_helpers.rb
@@ -1,0 +1,7 @@
+module UserHelpers
+
+  def stub_current_user_uuid(uuid)
+    allow_any_instance_of(ApplicationController).to receive(:current_user_uuid) { uuid }
+  end
+
+end
--- a/spec/support/user_helpers.rb
+++ b/spec/support/user_helpers.rb
@@ -1,7 +1,7 @@
 module UserHelpers
 
   def stub_current_user_uuid(uuid)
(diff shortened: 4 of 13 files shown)


app/controllers/api/v0/highlights_controller.rb
```diff
@@ -37,14 +39,12 @@ class Api::V0::HighlightsController < Api::V0::BaseController
...
 
   def create
-    binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
+    inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
     render(json: error, status: error.status_code) and return if error
 
-    # temporary test user_uuid.  will be removed when auth is set up
-    temp_user_uuid = '55783d49-7562-4576-a626-3b877557a21f'
-    created_highlight = Highlight.create!(binding.to_hash.merge(user_uuid: temp_user_uuid))
+    model = inbound_binding.create_model!(user_uuid: current_user_uuid)
```
steveburkett, line 45: seems odd to me to put the model factory in the binding.  it's much more common to see different createX factories in the model. that's where a rails dev would first think of going to find a create factory for a model. was there a specific reason to put this into the binding?

config/initializers/highlight_override.rb
```diff
@@ -53,3 +53,9 @@ def self.create_from_model(model)
     new(model.attributes)
   end
 end
+
+Api::V0::Bindings::NewHighlight.class_exec do
```
steveburkett, line 57: Wonder if this create_model! should be #create_highlight (it seems confusing that a NewHighlight creates a Highlight..just the naming seemed like it was perhaps a typo).  Also, is it necessary for this create_model! to have a bang? (the purpose of bang is to be used as a differentiator to a non-bang method, the more dangerous version of the non-bang version.  I believe it's not recommended to have a stand-along bang method.

spec/support/user_helpers.rb
```diff
@@ -0,0 +1,7 @@
+module UserHelpers
+
+  def stub_current_user_uuid(uuid)
+    allow_any_instance_of(ApplicationController).to receive(:current_user_uuid) { uuid }
```
steveburkett: allow_any_instance_of(ApplicationController).to receive(: current_user_uuid).and_return(: uuid)?  seems just a tad bit more intention revealing.  If there's a good reason to use the block form, would like to know that tho.

app/controllers/api/v0/highlights_controller.rb
steveburkett: i do see one of your reasons...in the previous form, was pulling out the binding's innards (the hash) in order to pass thru to HL.create.   so, yeah...maybe this form is a bit better OO.    still have the bang? tho (referenced below)

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -37,14 +39,12 @@ class Api::V0::HighlightsController < Api::V0::BaseController
   end
 
   def create
-    binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
+    inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
     render(json: error, status: error.status_code) and return if error
```
steveburkett, line 43: it might be good to run this one thru rubocop.  it's correct to use and, but i remember rubocup once telling me on and statements like this to use && instead of and.  (it might be clearer to put the return on the next line).   sometimes and statements get lost in longer lines

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,21 @@
 class ApplicationController < ActionController::API
 
+  protected
+
   include RescueFromUnlessLocal
 
+  def current_user_uuid
+    @current_user_uuid ||= begin
+      if Rails.env.development? && ENV['STUBBED_USER_UUID']
```
steveburkett, line 9: do we want a warning here if the STUBBED_USER_UUID is set, but not in development mode?  could this create confusion?  

another ?  is it a problem that this is memoized? if a user is removed or logs out of accounts while this user_uuid is still in use could this be a problem?

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,21 @@
...
+        ENV['STUBBED_USER_UUID']
+      else
+        OpenStax::Accounts::Sso.user_uuid(request)
+      end
+    end
+  end
+
+  def render_unauthorized_if_no_current_user
+    head :unauthorized if current_user_uuid.nil?
```
steveburkett, line 23: we dont want an error binding returned on this?  just the headers?

app/controllers/api/v0/highlights_controller.rb
jpslav: To me, the model is what other things depend on; it shouldn't have dependencies out to other things.  So I like that the bindings know about the model but not the other way around.  If the createX factory methods were in the model we'd eventually have `createFromV0Binding`, `createFromV1Binding`, etc, which I think you'd agree wouldn't be good.  For every binding version, we need a place to put code that goes from binding to model and model to binding.  The two options I see are putting that code in the binding or making a new mapper class for each binding (e.g. `Api::V0::BindingModelMappers::Highlight`).  I tend to prefer in the binding approach but I could go either way.  What do you think?

The bang is to reference that the model is being created with `create!` (to indicate that if errors happen they will raise exceptions instead of setting errors on the highlight model).

config/initializers/highlight_override.rb
jpslav: I could see devs reading `NewHighlight#create_highlight` as creating a Highlight binding instance (which is actually how I first read your statement "it seems confusing that a NewHighlight creates a Highlight").  `#create_model` implies to me making an AR model.  Should we go with `#create_highlight_model` to be clearer?

Like said above, the bang here for me is a way to tell the devs that this method calls `Highlight.create!` internally and to expect exceptions.  I know there's the argument about `!` only being used when there's a non-bang version that is less dangerous, but in this context I think it echoes the `save`/`save!` and `create`/`create!` activerecord methods which have never really seemed about danger but more about exceptions and performance.

spec/support/user_helpers.rb
jpslav: I can change this.  I normally go with the block form to make shorter lines :-)

app/controllers/api/v0/highlights_controller.rb
jpslav: rubocop does complain but they have different precedences and as avdi grimm points out they are more control flow operators than boolean logic operators: http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/

I don't really want to change this to

```ruby
if error
  render(json: error, status: error.status_code)
  return
end
```

app/controllers/application_controller.rb
jpslav: Sure re first comment.  I made the change, please check my logic.

Not a problem that it is memoized.  Controller instances are created anew for each request, so this memoized variable only lasts for the duration of one request.  The login state cannot change within a request because it is based on the cookie which doesn't change during the request (at least not to Rails' knowledge)

app/controllers/application_controller.rb
jpslav: The swagger docs don't have an error binding for the 401 error (only some of the error responses do): https://github.com/openstax/highlights-api/blob/master/app/controllers/api/v0/swagger_responses.rb#L4-L6

config/initializers/highlight_override.rb
jpslav: I was about to change it to `#create_highlight_model` but then I noticed that we already have `#create_from_model` so now I'm wondering if we should leave it so we have consistency in `#create_model` and `#create_from_model`.

app/controllers/api/v0/highlights_controller.rb
steveburkett: "If you have a method that is marked with a bang at its end, then a version without it should also exist." https://blog.codeminer42.com/and-understanding-one-of-rubys-coolest-naming-conventions-5a9300b75605

steveburkett: agree about the flow of dependency arrows.  but i wasnt arguing to put a dependency on the binding inside the model; i was suggesting a clean factory method inside the model that is passed the params.  or perhaps another abstraction.   that said..i think the binding factory is fine, if it's consistent.

config/initializers/highlight_override.rb
steveburkett: @jpslav agreed.  in the end..it's about consistency, readability.   these two are symmetrical and are at the same abstraction level, so sounds good.

app/controllers/api/v0/highlights_controller.rb
steveburkett: just curious...why do you not like the multi-line way?  it's crystal clear w/ no chance of operator precedence ambiguity at all.

app/controllers/api/v0/highlights_controller.rb
jpslav: Yeah I know that some people have that opinion about `!` I just don't subscribe to it so wholeheartedly :-)

app/controllers/api/v0/highlights_controller.rb
jpslav: too long.  `blah and return` is fairly common and I like the conciseness.

steveburkett (approved):

PR #14, opened 2019-11-07
Get highlights

steveburkett:
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/605

Note: that validation of source/source_parent_ids happens in the model, not swagger.

image


app/controllers/api/v0/highlights_controller.rb
```diff
@@ -38,6 +38,54 @@ class Api::V0::HighlightsController < Api::V0::BaseController
...
+      parameter do
+        key :name, :get_highlights
+        key :in, :query
+        key :description, 'The get highlight query parameters'
+        key :required, true
+        schema do
+          key :'$ref', :GetHighlights
+        end
+      end
+      response 201 do
```
jpslav: 200

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -38,6 +38,54 @@ class Api::V0::HighlightsController < Api::V0::BaseController
...
+      end
+      extend Api::V0::SwaggerResponses::AuthenticationErrorId
+      extend Api::V0::SwaggerResponses::UnprocessableEntityError
+      extend Api::V0::SwaggerResponses::ServerError
+    end
+  end
+  # Get highlights
+  # Example call:
+  #      /api/v0/highlights?source_type=”openstax_page”&source_parent_id=[“123”]&
+  #         color=#ff0000&page=1&per_page=20&order=desc
```
jpslav: Can you put this in the swagger endpoint description so visible to people looking at the docs?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -48,6 +48,56 @@ def create
...
+      key :operationId, 'getHighlights'
+      key :produces, [
+        'application/json'
+      ]
+      key :tags, [
+        'Highlights'
+      ]
+      # security do
+      #   key :api_id, []
+      # end
```
jpslav: Looks like [swagger 2 doesn't have cookie auth](https://swagger.io/docs/specification/2-0/authentication/) as an option (open API 3 adds it), so I think we can take these out?  Unless you see something to the contrary.

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -48,6 +48,56 @@ def create
     render json: response_binding, status: :created
   end
 
+  swagger_path '/highlights' do
+    operation :get do
+      key :summary, 'Get the highlight(s)'
```
jpslav: "Get selected highlights" probably more apt

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -48,6 +48,56 @@ def create
...
+  # Example call:
+  #      /api/v0/highlights?source_type=”openstax_page”&source_parent_id=[“123”]&
+  #         color=#ff0000&page=1&per_page=20&order=desc
+  def index
+    inbound_binding, error = bind(request.query_parameters, Api::V0::Bindings::GetHighlights)
+    render(json: error, status: error.status_code) and return if error
+
+    highlights = inbound_binding.query(user_uuid: current_user_uuid)
+
+    response_binding = Api::V0::Bindings::Highlights.create_from_model(highlights)
```
jpslav: pluralize `create_from_model` since input is multiple models?

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,69 @@ module Api::V0::Swagger::Models::Highlight
     end
   end
 
+  swagger_schema :GetHighlights do
+    key :required, [:source_type]
+    property :page do
+      key :type, :integer
+      key :description, 'The page number'
```
jpslav: Need to be more specific here somehow so people don't think "OpenStax page number".  E.g. "The page number of paginated results, zero-indexed".  

Specify the default? (guessing it is zero).

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,69 @@ module Api::V0::Swagger::Models::Highlight
...
+    end
+    property :per_page do
+      key :type, :integer
+      key :description, 'The number of highlights to return per page'
```
jpslav: Ditto, could be misread as "per OpenStax page".  "per paginated results page" would be clearer.

Also, is there a default here?  Is pagination on by default?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -48,6 +48,56 @@ def create
     render json: response_binding, status: :created
   end
 
+  swagger_path '/highlights' do
+    operation :get do
+      key :summary, 'Get the highlight(s)'
+      key :description, 'Get the highlight(s) for the owner.'
```
jpslav: You may want something multiline here, to say that results can be paginated, filtered by various attributes, ordered variously, etc.

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,69 @@ module Api::V0::Swagger::Models::Highlight
...
+    end
+    property :order do
+      key :type, :string
+      key :enum, %w[asc desc]
+      key :description, 'The returned sorted order'
```
jpslav: Say the order is based on whatever fields those are in the Highlight.

Specify the default if there is one.

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,69 @@ module Api::V0::Swagger::Models::Highlight
...
   end
 
+  swagger_schema :Highlights do
+    property :highlights do
+      key :type, :array
+      key :description, 'One or more highlights'
+      items do
+        key :'$ref', :Highlight
+      end
+    end
```
jpslav, line 108: return total count, current page, per page?

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,69 @@ module Api::V0::Swagger::Models::Highlight
...
+  end
+
   swagger_schema :NewHighlight do
     key :required, [:source_type, :source_id, :anchor, :highlighted_content, :color, :location_strategies]
   end
 
+  add_properties(:GetHighlights, :NewHighlight, :Highlight) do
```
jpslav: Not sure if `GetHighlights` should share these properties, because the descriptions need to be different.  E.g. for `:source_parent_ids` in `Highlight` and `NewHighlight`, the field is for all of the parent IDs of the highlight whereas in `GetHighlight` the description should be something like "One or more unrelated parent IDs; query results will have at least one parent ID that matches those provided"

Ditto the others a little, e.g. "The source_type of the highlight" vs "The source_type to limit results to"

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,69 @@ module Api::V0::Swagger::Models::Highlight
...
+    property :source_type do
+      key :type, :string
+      key :enum, ['openstax_page']
+      key :description, 'The source_type of the highlight, typically a openstax_page'
```
jpslav: Probably shouldn't use "source_type" in the description of "source_type".  Maybe just "The type of content that contains the highlight, typically..."

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,69 @@ module Api::V0::Swagger::Models::Highlight
...
+    end
+    property :source_parent_ids do
+      key :type, :array
+      key :description, 'The parent IDs of the highlight. For book highlights, ' \
+                        'the parent_id could be a book, unit, or chapter ID'
```
jpslav: "parent IDs of the highlight" isn't quite right.  sounds like parents of the highlights when really it is parents of the source the highlight is in.  "IDs of the parent sources of the source containing the highlight.  E.g. For an openstax_page highlight, the IDs would be those of the chapter, unit, and book.

Also note that it should be "and" instead of "or"

Does the order of these IDs matter in the array?  If not say that.

app/models/highlight.rb
```diff
@@ -19,6 +19,14 @@ class Highlight < ApplicationRecord
 
   before_validation :normalize_color, :normalize_source_ids
 
+  scope :by_source_type, ->(source_type) { where(source_type: source_type) }
+  scope :by_color, ->(color) { where(color: color) }
+  scope :by_user, ->(user_id) { where(user_uuid: user_id) }
+  scope :by_order, ->(order_direction) { order(created_at: order_direction.to_sym) }
```
jpslav: If this is pagination order it should be based on the two highlight order fields.

app/models/highlight.rb
```diff
@@ -19,6 +19,14 @@ class Highlight < ApplicationRecord
...
+  scope :by_source_parent_ids, ->(source_parent_ids) do
```
jpslav, line 32: Some way to clarify that just one of the parent IDs needs to match? `by_any_source_parent_id`?

app/models/highlight.rb
```diff
@@ -29,4 +37,10 @@ def normalize_source_ids
       source_parent_ids.map!(&:downcase)
```
jpslav: do we need to trim/strip the IDs too?

jpslav: Also where do we document that for openstax_page the source IDs have such and such form?  Do we need to check for that form?  E.g. is `@12` version stuff included?

config/initializers/highlight_override.rb
```diff
@@ -32,9 +34,7 @@ def list_invalid_properties
       return invalid_properties
     end
 
-    if location_strategies.any?(&:nil?)
-      invalid_properties.push('invalid strategy detected')
-    end
+    invalid_properties.push('invalid strategy detected') if location_strategies.any?(&:nil?)
```
jpslav: more specific error like "no location strategies provided"?

actually note that this check is already done in the generated `NewHighlight` binding, so leaving this would leave a duplicate error.

config/initializers/highlight_override.rb
```diff
@@ -59,3 +58,26 @@ def create_model!(user_uuid:)
...
+
+    # The submitted GetHighlight properties create automatic chaining via
+    # the by_X scopes on the Highlight model.
+    pagination = to_hash.dup
+    filter_by = pagination.slice!(:page, :per_page)
+    filter_by.each do |key, value|
+      highlights = highlights.public_send("by_#{key}", value) if value.present?
+    end
+
+    highlights.paginate(pagination)
```
jpslav: This blows up if `:page` is not provided.  They are not required, so need to be clear about defaults or not paginate if pagination parameters not provided.  Also I think if pagination is specified we need to also have a default order?  I'm not a fan of default scopes for order defaults fwiw -- always seems to be problematic at some point -- would be happier with a default here if pagination used.

db/seeds.rb
```diff
@@ -1,7 +1,27 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+Highlight.destroy_all
```
jpslav: thought we were using factories instead of seeds?

spec/factories/highlights.rb
```diff
@@ -9,13 +9,14 @@
...
+          type: 'TextPositionSelector',
           start: 12,
           end: 10
-         }
-     ].to_json
+        }
+      ].to_json
     end
     source_id { SecureRandom.uuid }
+    source_parent_ids { [123, 456] }
```
jpslav: I think the schema says this is an array of strings, not an array of integers.

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -5,6 +5,63 @@
 
   before { allow(Rails.application.config).to receive(:consider_all_requests_local) { false } }
 
+  describe 'GET /highlights' do
+    let!(:highlight1) { create(:highlight, user_uuid: user_uuid) }
+    let!(:highlight2) { create(:highlight, user_uuid: user_uuid) }
+    let!(:highlight3) { create(:highlight) }
```
jpslav, line 15: btw, you don't necessarily need to use the bang version of `let!` -- I only used it in that other spec b/c if I didn't my `expect{blah}.to change{Highlight.count}.by(1)` got messed up because the instantiation happened in the `blah` part.

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -5,6 +5,63 @@
...
+
+    let(:per_page) { 10 }
```
jpslav, line 18: should have some specs that don't set this or `page` and that will show you the kaboom

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -5,6 +5,63 @@
...
+    let(:query_params) do
+      {
+        source_type: 'openstax_page',
+        source_parent_ids: highlight1.source_parent_ids,
```
jpslav: should test with not all the source parent IDs as clients are likely to call with just one.

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -48,6 +48,56 @@ def create
...
+        end
+      end
+      response 201 do
+        key :description, 'Returns the desired highlight(s).'
+        schema do
+          key :'$ref', :Highlights
+        end
+      end
+      extend Api::V0::SwaggerResponses::AuthenticationError
+      extend Api::V0::SwaggerResponses::ForbiddenError
```
jpslav: Forbidden (403) not possible since we aren't querying specific highlights belonging to someone else, just those belonging to the user.

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -5,6 +5,63 @@
...
+      end
+
+      it 'gets the highlights for the user' do
+        get highlights_path, params: query_params
+        expect(response).to have_http_status(:ok)
+
+        highlights_json = json_response[:highlights]
+
+        expect(highlights_json.count).to eq 2
+        expect(highlights_json.first[:user_uuid]).to eq user_uuid
```
jpslav: to be paranoid you could check all returned user uuids

```ruby
expect(highlights_json.map{|json| json[:user_uuid]}.uniq).to eq [user_uuid]
```

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -5,6 +5,63 @@
...
+    context 'when the user is logged in' do
+      before do
+        stub_current_user_uuid(user_uuid)
+      end
+
+      it 'gets the highlights for the user' do
+        get highlights_path, params: query_params
+        expect(response).to have_http_status(:ok)
+
+        highlights_json = json_response[:highlights]
```
jpslav: is this still json? I think it is just an array of hashes, so `highlights_array` or `highlights` might be better

jpslav (changes requested):
Looks nice.  Probably needs some more specs (testing different query combinations, pagination options, defaults, etc)

app/controllers/api/v0/highlights_controller.rb
steveburkett: good point!

spec/requests/api/v0/highlights_controller_spec.rb
steveburkett: it was json before json_response converted it :)  will rename to highlights (more rubyish)

db/seeds.rb
steveburkett: I was using the seeds because i use Postman to query the API a lot.  Do you want me to delete the seeds?  it's not necessary for tests or anything, as factories are.  I just use them for convenience for hitting the API.

spec/requests/api/v0/highlights_controller_spec.rb
steveburkett: this set of tests (mostly) depends on these 3 test records present in the db, before the tests run.   the let! makes sure they run before the tests.

config/initializers/highlight_override.rb
steveburkett: i added in defaults for pagination.

app/models/highlight.rb
steveburkett: need to ask about this...i had wondered this myself.

app/models/highlight.rb
steveburkett: personally, i thought this seemed fine.    right now, you can see the trick i do to build the filter dynamically 

```
    filter_by.each do |key, value|
      highlights = highlights.public_send("by_#{key}", value) if value.present?
    end
```

so if we want this to be by_any_source_parent_ids, it would mean changing the swagger binding to property :any_source_parent_id, so that this by_x trick works.

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -48,6 +45,62 @@ def create
     render json: response_binding, status: :created
   end
 
+  swagger_path '/highlights' do
+    operation :get do
+      key :summary, 'Get selected highlights'
+      key :description, <<~DESC
+        Get selected highlight(s) for the owner.
```
jpslav: "for the owner" --> "belonging to the calling user"

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -48,6 +45,62 @@ def create
...
+
+        Highlights can be filtered thru query parameters:  source_type,
+        source_parent_ids, and color.
+
+        Highlights are always paginated.  Pass in page and per_page to
+        request specific page, or omit the pagination params to get the defaults
+        of page = 1 and per_page of 15.
```
jpslav, line 59: Can this be `page = #{Api::V0::Bindings::GetHighlights::PAGING_DEFAULTS[:page]}` and similar for `per_page`?

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,84 @@ module Api::V0::Swagger::Models::Highlight
     end
   end
 
+  swagger_schema :GetHighlights do
+    key :required, [:source_type]
+    property :page do
+      key :type, :integer
+      key :description, 'The page number of paginated results, one-indexed. Defaults to 1.'
```
jpslav, line 49: Ditto above on not repeating these defaults in literal strings - refer to some variable.

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,84 @@ module Api::V0::Swagger::Models::Highlight
...
+    end
+    property :per_page do
+      key :type, :integer
+      key :description, 'The number of highlights per page for paginated results. Defaults to 15.'
+    end
+    property :order do
+      key :type, :string
+      key :enum, %w[asc desc]
+      key :description, 'The desired sort order on the highlight creation time. Defaults to asc.'
```
jpslav: not on highlight creation time

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -42,25 +42,84 @@ module Api::V0::Swagger::Models::Highlight
...
+        key :type, :integer
+        key :description, 'The response page number'
+      end
+      property :per_page do
+        key :type, :integer
+        key :description, 'The response per page'
+      end
+      property :total_count do
+        key :type, :integer
+        key :description, 'The number of returned highlights'
```
jpslav: think this is intended to be the number of total results, not just those in this page of results (which are what is "returned")

app/models/highlight.rb
jpslav: ok, it is fine just make sure to make it clear in the description for this field in `GetHighlights` (which requires splitting that definition out from `Highlight` and `NewHighlight`

config/initializers/highlight_override.rb
```diff
@@ -59,3 +58,40 @@ def create_model!(user_uuid:)
     Highlight.create!(to_hash.merge(user_uuid: user_uuid))
   end
 end
+
+Api::V0::Bindings::Highlights.class_exec do
+  def self.create_from_models(highlights, pagination)
+    highlights = highlights.to_a unless highlights.is_a? Array
+    meta = pagination.merge(total_count: highlights.count)
```
jpslav: don't think this is what total count means

jpslav (changes requested):

app/controllers/api/v0/highlights_controller.rb
steveburkett: added to the TODO in this PR

config/initializers/highlight_override.rb
steveburkett: i had thought total_count meant the number returned in the response.  i changed it to return the total count of all highlights involved in all pages.

app/models/highlight.rb
steveburkett: added a custom validation for source_parent_ids to the Highlight model.

config/initializers/highlight_override.rb
steveburkett: i fixed the total count to be everything.

config/initializers/highlight_override.rb
```diff
@@ -60,9 +60,9 @@ def create_model!(user_uuid:)
 end
 
 Api::V0::Bindings::Highlights.class_exec do
-  def self.create_from_models(highlights, pagination)
+  def self.create_from_models(highlights, pagination, total_count)
     highlights = highlights.to_a unless highlights.is_a? Array
-    meta = pagination.merge(total_count: highlights.count)
+    meta = pagination.merge(total_count: total_count)
     attribs = { meta: meta, data: highlights }
     highlights_response = new(attribs)
```
jpslav: The `highlights` passed in here is an array of ActiveRecords.  Needs to be an array of `Api::V0::Binding::Highlight`.

steveburkett: fixed.

PR #16, opened 2019-11-11
Add Scout and fix some sentry issues

jpslav:
Add Scout and fix some sentry issues

Scout reporting:

image

Sentry reporting:

image

### DIFF
--- a/lib/rescue_from_unless_local.rb
+++ b/lib/rescue_from_unless_local.rb
@@ -12,7 +12,9 @@ module RescueFromUnlessLocal
     # exceptions (good in development) or rescue exceptions (good in production).  Because
     # the test of the setting is done at the moment the exception is being handled, we
     # can change the value in specs to test both the exception and the rescuing behavior.
-    def rescue_from_unless_local(*klasses, with: nil, &block)
+    # If send_to_sentry is true and the exception is not already going to be reraised,
+    # the exception will be captured off to Sentry.
+    def rescue_from_unless_local(*klasses, with: nil, send_to_sentry: false, &block)
       rescue_from(*klasses, with: with) do |exception|
         # This new_block assignment borrowed from pieces of ActiveSupport::Rescuable
         # The Symbol and Proc values of `with` have not been tested :-o
@@ -35,7 +37,12 @@ module RescueFromUnlessLocal
             end
           end
 
-        raise exception if Rails.application.config.consider_all_requests_local
+        if Rails.application.config.consider_all_requests_local
+          raise exception
+        elsif send_to_sentry
+          Raven.capture_exception(ex)
+        end
+
         instance_exec exception, &new_block
       end
     end
--- a/lib/rescue_from_unless_local.rb
+++ b/lib/rescue_from_unless_local.rb
@@ -40,7 +40,7 @@ module RescueFromUnlessLocal
         if Rails.application.config.consider_all_requests_local
           raise exception
         elsif send_to_sentry
-          Raven.capture_exception(ex)
+          Raven.capture_exception(exception)
         end
 
         instance_exec exception, &new_block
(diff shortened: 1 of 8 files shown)


lib/rescue_from_unless_local.rb
```diff
@@ -35,7 +37,12 @@ def rescue_from_unless_local(*klasses, with: nil, &block)
             end
           end
 
-        raise exception if Rails.application.config.consider_all_requests_local
+        if Rails.application.config.consider_all_requests_local
```
steveburkett, line 40: we should prob. have a unit test for this file that makes sure the right messages are sent out (to Raven) if sentry is turned on.

lib/tasks/install_secrets.rake
```diff
@@ -45,26 +45,40 @@ task :install_secrets, [] do
     end
   end
 
-  database_secrets = secrets.delete('database')
+  database_secrets = secrets.delete(:database)
+  write_yaml_file("config/database.yml", {
```
steveburkett, line 49: seems like this sort of thing should go into a gem called "openstax-rails" or something...it's something all rails apps at openstax need to do for boilerplate eh?

lib/tasks/install_secrets.rake
```diff
@@ -45,26 +45,40 @@ task :install_secrets, [] do
...
+    production: {
+      database: database_secrets[:name],
+      host: database_secrets[:url],
+      port: database_secrets[:port],
+      username: database_secrets[:username],
+      password: database_secrets[:password],
+      adapter: "postgresql",
+      encoding: "unicode",
+      pool: '<%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>'
```
steveburkett: does 5 match what the needs of all puma's threads and workers? (curious if 5 has been reasoned out)

lib/tasks/install_secrets.rake
```diff
@@ -45,26 +45,40 @@ task :install_secrets, [] do
...
+  })
+end
 
-  File.open(File.expand_path("config/secrets.yml"), "w") do |file|
-    # write the secrets hash as yaml, getting rid of the "---\n" at the front
-    file.write({'production' => secrets}.to_yaml[4..-1])
+def write_yaml_file(filename, hash)
+  File.open(File.expand_path(filename), "w") do |file|
+    # write the hash as yaml, getting rid of the "---\n" at the front
+    file.write(hash.deep_stringify_keys.to_yaml[4..-1])
```
steveburkett, line 81: would something like `yaml.gsub("---\n", '')` be better than these 2 lines?  it's one line and self documenting?

steveburkett (approved):

lib/tasks/install_secrets.rake
jpslav: I would conceivable put the `write_yaml_file` in a gem, but not the specific lines here

lib/tasks/install_secrets.rake
jpslav: good question.  I have no idea.  can I give one to you on the backburner?

lib/tasks/install_secrets.rake
jpslav: good point.  we should do that once we move it somewhere and have a spec for it.

lib/tasks/install_secrets.rake
jpslav: >If you are using the Puma web server we recommend setting the pool value to equal ENV['RAILS_MAX_THREADS']

https://devcenter.heroku.com/articles/concurrency-and-database-connections
so it should be 10.

lib/rescue_from_unless_local.rb
jpslav: agreed, but deferring to moving this into a gem (https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/673)

lib/tasks/install_secrets.rake
jpslav: delayed to https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/673

PR #17, opened 2019-11-11
highlight deletion endpoint

jpslav:
highlight deletion endpoint

### DIFF
--- a/app/controllers/api/v0/highlights_controller.rb
+++ b/app/controllers/api/v0/highlights_controller.rb
@@ -1,7 +1,7 @@
 class Api::V0::HighlightsController < Api::V0::BaseController
   before_action :render_unauthorized_if_no_current_user
 
-  before_action :get_highlight, only: [:show, :update, :destroy]
+  before_action :get_highlight, only: [:update, :destroy]
 
   swagger_path '/highlights' do
     operation :post do
--- a/app/controllers/api/v0/highlights_controller.rb
+++ b/app/controllers/api/v0/highlights_controller.rb
@@ -1,7 +1,7 @@
 class Api::V0::HighlightsController < Api::V0::BaseController
   before_action :render_unauthorized_if_no_current_user
 
-  # before_action :set_highlight, only: [:show, :update, :destroy]
+  before_action :get_highlight, only: [:show, :update, :destroy]
 
   swagger_path '/highlights' do
     operation :post do
@@ -32,7 +32,7 @@ class Api::V0::HighlightsController < Api::V0::BaseController
           key :'$ref', :Highlight
         end
       end
-      extend Api::V0::SwaggerResponses::AuthenticationErrorId
+      extend Api::V0::SwaggerResponses::AuthenticationError
       extend Api::V0::SwaggerResponses::UnprocessableEntityError
       extend Api::V0::SwaggerResponses::ServerError
     end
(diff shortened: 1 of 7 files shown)


app/controllers/api/v0/highlights_controller.rb
```diff
@@ -48,9 +48,45 @@ def create
...
+    @highlight.destroy!
+    head :ok
+  end
+
   private
 
-  def set_highlight
+  def get_highlight
     @highlight = Highlight.find(params[:id])
+    raise SecurityTransgression unless @highlight.user_uuid == current_user_uuid
```
steveburkett, line 89: is there any use case of an admin who can delete other's highlights?

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -98,4 +100,62 @@
       end
     end
   end
+
+  describe 'DELETE /highlights/{id}' do
+    let!(:highlight) { FactoryBot.create(:highlight) }
```
steveburkett, line 105: personal thing.  i generally like stuff like this done in a before block instead of a let!.  the former seems more intentional, the latter seems a bit more hacky (and i've known lots of devs to miss seeing the !, esp when it gets mixed in w/ other lets)

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -98,4 +100,62 @@
...
+
+    context 'when the user is not logged in' do
+      context 'when the highlight does not exist' do
+        it 'gives a 401 before a 404' do
```
steveburkett, line 109: in this specific test case, you dont need to do the factory create right?  the 401 happens first before even needing the HL to be written to the db?

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -98,4 +100,62 @@
...
+      it 'is deleted' do
+        expect{
+          delete highlights_path(id: highlight.id)
+        }.to change{Highlight.count}.by(-1)
+
+        expect(response).to have_http_status :ok
+      end
+
+      context 'when the highlight does not exist' do
+        it '404s' do
```
steveburkett, line 138: 404s?  how about it 'returns not_found'?  (no biggie, minor question)

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -98,4 +100,62 @@
...
+          delete highlights_path(id: highlight.id)
+        }.not_to change{Highlight.count}
+
+        expect(response).to have_http_status(:forbidden)
+      end
+    end
+  end
+
+  def highlights_path(id: nil)
+    "/api/v0/highlights#{id.nil? ? '' : "/#{id}"}"
```
steveburkett: should the version be in the params, defaulted?

steveburkett (approved):

app/controllers/api/v0/highlights_controller.rb
jpslav: no, figured if we need that we'll just do it on the console

spec/requests/api/v0/highlights_controller_spec.rb
jpslav: yes that is true in this case, I'm ok making the DB hit here vs rearranging the contexts to use the `let!` only where needed, I think

PR #18, opened 2019-11-13
Highlight update endpoint

jpslav:
Highlight update endpoint


app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -114,6 +116,19 @@ module Api::V0::Swagger::Models::Highlight
     end
   end
 
+  add_properties(:HighlightUpdate) do
```
steveburkett: can we keep the same pattern for others?  :UpdateHighlight (symmetrical to NewHighlight)

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -114,6 +116,19 @@ module Api::V0::Swagger::Models::Highlight
...
+    property :color do
+      key :type, :string
+      # remove the anchors because swagger-codegen always escapes them
+      key :pattern, ::Highlight::VALID_COLOR.inspect[1..-2]
```
steveburkett: maybe we should store this "concept" on the model, so we can refer to one place for it.

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -114,6 +116,19 @@ module Api::V0::Swagger::Models::Highlight
...
+      key :description, 'The highlight color.'
```
steveburkett, line 140: should this say something a bit more specific like 'The highlighted color to update to'?

app/controllers/api/v0/swagger/models/highlight.rb
```diff
@@ -114,6 +116,19 @@ module Api::V0::Swagger::Models::Highlight
...
+    end
+    property :annotation do
```
steveburkett: can you send in an update with neither of these properties? (e.g., one is required?) . or are both required?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -101,6 +101,56 @@ def index
...
+        end
+      end
+      extend Api::V0::SwaggerResponses::AuthenticationError
+      extend Api::V0::SwaggerResponses::UnprocessableEntityError
+      extend Api::V0::SwaggerResponses::ServerError
+    end
+  end
+
+  def update
+    inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::HighlightUpdate)
```
steveburkett: do we want to rails log that an update happened?  anything that records what happened?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -101,6 +101,56 @@ def index
...
+    render(json: error, status: error.status_code) and return if error
+
+    model = inbound_binding.update_model!(@highlight)
+
+    response_binding = Api::V0::Bindings::Highlight.create_from_model(model)
```
steveburkett, line 151: when we update, do we want to return a different json format:   before_data and after_data (changed)?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -101,6 +101,56 @@ def index
...
+      key :produces, [
+        'application/json'
+      ]
+      key :tags, [
+        'Highlights'
+      ]
+      parameter do
+        key :name, :id
+        key :in, :path
+        key :description, 'ID of the highlight to update'
```
steveburkett: needs a period

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -101,6 +101,56 @@ def index
...
+        key :required, true
+        key :type, :string
+        key :format, 'uuid'
+      end
+      parameter do
+        key :name, :highlight
+        key :in, :body
+        key :description, 'The highlight updates'
```
steveburkett: needs a period

steveburkett:
would be nice to see a screenshot of a curl or something that did an update from outside of the specs.

app/controllers/api/v0/swagger/models/highlight.rb
jpslav: I had named it like this b/c it isn't a full Highlight, it is just an update to a highlight... but I think I could get to the other way of thinking that the name is "VerbHighlight"

app/controllers/api/v0/swagger/models/highlight.rb
jpslav: Can you elaborate?

app/controllers/api/v0/swagger/models/highlight.rb
jpslav: Changed to "The new highlight color."

app/controllers/api/v0/swagger/models/highlight.rb
jpslav: Yes you can, in which case nothing will change.

app/controllers/api/v0/highlights_controller.rb
jpslav: We'll eventually be sending events to Kafka.  Will be more useful there than in the log.

app/controllers/api/v0/highlights_controller.rb
jpslav: I don't think so.  Unless you have some specific use case for that in mind.  The client could have the `before_data` if it needs it.

jpslav:
Re https://github.com/openstax/highlights-api/pull/18#issuecomment-553642689 I'd rather there be a selenium / capybara spec that does this more "real-life" test b/c a one-off curl spec doesn't give us long-term mileage.  Made a card for this https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/682  @steveburkett

app/controllers/api/v0/swagger/models/highlight.rb
steveburkett: e.g, as `::Highlight::VALID_COLOR_WOUT_ANCHORS`

steveburkett (approved):

PR #20, opened 2019-11-16
drop paging, switch from parent IDs to scope ID

jpslav:
* drop paging
* switch from parent IDs to scope ID
* Add linked list ordering (have to specify prev and next highlights on create, those used to provide order to queries)


app/models/highlight.rb
```diff
@@ -16,54 +21,113 @@ class Highlight < ApplicationRecord
...
-      source_parent_ids.map!(&:downcase).map!(&:strip)
+      self.scope_id = scope_id&.downcase&.strip
     end
   end
 
-  private
+  protected
+
+  def all_from_scope_and_source
+    Highlight.where(scope_id: scope_id).where(source_id: source_id)
```
steveburkett, line 52: perhaps this can be built reusing the scopes created above?  or to create a by_scope_source_id scope that uses both?  (maybe the chained version might be reused sometime)

app/models/highlight.rb
```diff
@@ -16,54 +21,113 @@ class Highlight < ApplicationRecord
...
-    return unless openstax_page?
+  def validate_uuid(field, is_source_type_specific:)
+    unless VALID_UUID.match?(send(field).to_s.downcase)
+      errors.add(field, "must be a UUID#{' for source_type ' + source_type if is_source_type_specific}")
+    end
+  end
+
+  def must_have_source_type_specific_ids
+    if openstax_page?
+      validate_uuid(:scope_id, is_source_type_specific: true)
```
steveburkett: this seems a bit confusing.  would we ever validate a UUID when not source_type specific?  this is already inside the if openstax_page?, so it's just validate_uuid right? later on... we'd have a `validate_id` or something right?

app/models/highlight.rb
```diff
@@ -16,54 +21,113 @@ class Highlight < ApplicationRecord
...
-    order_direction = order_direction.to_sym
-    order(order_in_source: order_direction, source_order: order_direction)
-  }
-  scope :by_source_parent_ids, ->(source_parent_ids) do
-    where('source_parent_ids && ?', postgres_style_array(source_parent_ids))
-  end
+  scope :by_scope_id, ->(scope_id) { where(scope_id: scope_id) }
+  scope :by_source_ids, ->(source_ids) { where(source_id: source_ids) }
+
+  around_create :insert_new_highlight_between_neighbors
```
steveburkett, line 35: we've already discussed this and i know your opinion, just restating my opinion here for the record.   it's really good to avoid AR callbacks http://codeatmorning.com/3-reasons-to-avoid-rails-callbacks/.  This app will grow, tests will continue to be added, more model associations will be added....these sorts of hooks are 99% guaranteed to cause headaches in the future.    

A good alternative to them, imo, is to create a higher level abstraction:
```
     HighlightNeighborConnector
        def create 
            set_neighbors(highlight)
            highlight.create
       end

      def destroy(highlight)
          reconnect_neighbors
          highlight.destroy
     end
```

steveburkett:
The endContainer needs to be respelled to end_container

app/models/highlight.rb
```diff
@@ -16,54 +21,113 @@ class Highlight < ApplicationRecord
...
+    if prev_highlight_id.nil? && next_highlight_id.nil? && all_from_scope_and_source.any?
+      errors.add(:base, 'Must specify previous or next highlight because there are other highlights ' \
+                        'in this scope and source.')
+    end
 
-    source_parent_ids.each do |source_parent_id|
-      unless valid_uuid?(source_parent_id)
-        errors.add(:source_parent_ids, 'source parent id of type openstax page must be a uuid')
+    if prev_highlight_id
+      if prev_highlight.nil?
```
steveburkett: Would this "big" nested conditional be better to break into two concerns? (right now, it's responsible for 2: next and prev neighbors).   I ask this because of SRP, and because it looks to have repeated logic that could be generalized a bit. 

Perhaps must_have_consistent_neighbors could be

```
def must_have_consistent_neighbors
    validate_prev_neighbor
    validate_next_neighbor
end 
```

app/models/highlight.rb
```diff
@@ -16,54 +21,113 @@ class Highlight < ApplicationRecord
...
+        if next_highlight.scope_id != scope_id
+          errors.add(:next_highlight_id, 'does not live in the same scope')
+        end
+      end
     end
   end
 
-  def self.postgres_style_array(souce_parent_ids)
-    "{#{souce_parent_ids.join(',')}}"
+  def insert_new_highlight_between_neighbors
```
steveburkett, line 111: My vote would have been that this neighboring logic would be abstracted to a different class (where it could be tested independently of the model, sans the AR hooks).  Ignore this if you want P)

config/initializers/highlight_override.rb
```diff
@@ -60,39 +60,39 @@ def create_model!(user_uuid:)
...
+    to_hash.each do |key, value|
       highlights = highlights.public_send("by_#{key}", value) if value.present?
     end
-    pagination = PAGING_DEFAULTS.merge(pagination)
 
-    [highlights.paginate(pagination), pagination, highlights.count]
+    highlights = highlights.to_a
+
+    if source_ids.present?
+      # Sort the highlights in Ruby, not Postgres
```
steveburkett, line 87: seems like this comment is unnecessary...it's obvious from the code

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -147,18 +127,46 @@
...
+          end
+
+          it 'is owned by the right user' do
+            expect(Highlight.first.user_uuid).to eq user_uuid
+          end
         end
 
-        it 'is owned by the right user' do
-          expect(Highlight.first.user_uuid).to eq user_uuid
+        context 'when there is one highlight' do
```
steveburkett: perhaps this should be suffixed with 'already existing', to be clearer.

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -147,18 +127,46 @@
...
+          # Lots more invalid data tests in the model spec
+          it '422s when the new highlight does not specify prev or next' do
+            post highlights_path, params: valid_attributes
+            expect(response).to have_http_status(:unprocessable_entity)
+            expect(response.body).to match(/Must specify previous or next highlight/)
+          end
+
+          it 'puts the new highlight after the first one when set a prev highlight' do
```
steveburkett, line 157: does this need to explicitly test hl1 and new HLs' prev and next ids are all correct?

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -250,6 +258,26 @@
           expect(response).to have_http_status :not_found
         end
       end
+
+      context 'when there are two highlights' do
```
steveburkett, line 265: should there be other variations of the tests:   no neighbors, one neighbor prev, one neighbor after, 2 neighbors?

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -4,30 +4,33 @@
...
-    let!(:highlight2) { create(:highlight, user_uuid: user_uuid) }
-    let!(:highlight3) { create(:highlight) }
-
-    let(:page) { 1 }
-    let(:per_page) { 10 }
-    let(:source_parent_ids) { highlight1.source_parent_ids }
+    let!(:highlight1) { create(:highlight, user_uuid: user_uuid, source_id: source_id, scope_id: scope_1_id) }
+    let!(:highlight2) { create(:highlight, user_uuid: user_uuid,                       scope_id: scope_1_id) }
+    let!(:highlight3) { create(:highlight, user_uuid: user_uuid,                       scope_id: scope_1_id) }
+    let!(:highlight4) { create(:highlight, user_uuid: user_uuid, source_id: source_id, scope_id: scope_1_id, prev_highlight: highlight1) }
```
steveburkett, line 18: wonder if these should be renamed to better names to describe the utility?  e.g, highlight_with_prev_neighbor, highlight_with_both_neighbors, highlight_with_no_neighbors

app/models/highlight.rb
```diff
@@ -16,54 +21,113 @@ class Highlight < ApplicationRecord
...
+    self.order_in_source =
+      if prev_highlight && next_highlight
```
steveburkett, line 121: would be nice if this logic were in a method with a good name that describes what setting this float means.  Perhaps  `def find_relative_order_by_source` and with a short blurb on what would happen if the order_in_source reaches the mathematical end of the float range (if that's realistic). 

(This will definitely be one of those things that will really mess things up if anyone deletes records from postgres and not thru the model)

app/models/highlight.rb
```diff
@@ -16,54 +21,113 @@ class Highlight < ApplicationRecord
...
+        (prev_highlight.order_in_source + next_highlight.order_in_source)/2.0
+      elsif prev_highlight
+        prev_highlight.order_in_source + 1.0
+      elsif next_highlight
+        next_highlight.order_in_source - 1.0
+      else
+        0.0
+      end
+
+    yield # the create call, lets us have `id` set on next lines
```
steveburkett: What will be the "big O" performance impact of this when this table has a million+ rows?

app/models/highlight.rb
jpslav: I generally agree with you which is why I wrote https://github.com/lml/lev#routines (a routine would be like the above except with transactions built in).  The exception in my mind is when a callback only works inside one model.  Mostly I'm ok with this with a model affecting its own state.  I admit that this `Highlight` case is a little different because it is impacting its own state and the state of other highlight models.  So it is borderline for me, in this case I'm ok with it b/c I see Highlight as a linked list model so having the glue in there for linked listing is ok

app/models/highlight.rb
jpslav: I'm ok with this because it is pretty simple and hidden internally.  I do like the `by_blah` scopes as a public mechanism that helps with querying generically.

app/models/highlight.rb
jpslav: Makes sense.  Changed.

config/initializers/highlight_override.rb
jpslav: Haha I added this specifically because I thought you'd ask for it :-D

app/models/highlight.rb
jpslav: Moved it

app/models/highlight.rb
jpslav: O(3)? Create self, update prev, update next.  Are you seeing some other performance impact tied to the size of the table?

spec/requests/api/v0/highlights_controller_spec.rb
jpslav: I don't know if I could handle those line lengths ;-) I'm ok as is but if you want to change in a future PR that'd be ok by me.

config/initializers/highlight_override.rb
steveburkett: nah...comments in code are almost always a code smell.  except for e.g., comments to external links....

spec/requests/api/v0/highlights_controller_spec.rb
jpslav: There are a bunch of these variations on the model spec.

app/models/highlight.rb
jpslav: Did a change like this

spec/requests/api/v0/highlights_controller_spec.rb
jpslav: good point, I added one case of this in this spec and then more in the model spec where they were missing.

steveburkett (approved):

app/models/highlight.rb
steveburkett: Not concerned...i do see the indexes on update prev and update next.

PR #21, opened 2019-11-21
Add paging back in

jpslav:
Add paging back in with default page 1, default per page 15, max per page 200.


config/initializers/highlight_override.rb
```diff
@@ -90,9 +99,23 @@ def query(user_uuid:)
       end
 
       highlights.sort_by!{ |highlight| [source_id_order[highlight.source_id], highlight.order_in_source] }
+    else
+      # Have to sort by something for pagination to be sensible, choose created_at
+      highlights.sort_by!(&:created_at)
```
steveburkett, line 103: shouldnt the highlights be sorted by their linked list order?

config/initializers/highlight_override.rb
```diff
@@ -90,9 +99,23 @@ def query(user_uuid:)
...
     end
 
-    highlights
+    local_page = page || Api::V0::Swagger::Models::Highlight::DEFAULT_HIGHLIGHTS_PAGE
+    local_per_page = per_page || Api::V0::Swagger::Models::Highlight::DEFAULT_HIGHLIGHTS_PER_PAGE
+
+    first_index_in_page = (local_page - 1) * local_per_page
```
steveburkett: we could use will_paginate on ruby results too (not just models).  that could help sweep away the paging logic out of the model code.

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -45,24 +43,98 @@
...
+            expect(response).to have_http_status(:ok)
+            expect(highlights.map{|hl| hl[:id]}).to eq(
+              [highlight3, highlight2, highlight1, highlight5, highlight4].map(&:id)
+            )
+          end
+        end
+
+        context 'with paging' do
+          it('gets the user\'s highlights in the order of the source IDs and then by order within page and paginates') do
+            get highlights_path, params: query_params.merge(
```
steveburkett: this is one giant test.   perhaps broken up into a test per page?

config/initializers/highlight_override.rb
jpslav: No because without source IDs specified, the pages are going to be all mixed up anyway.

config/initializers/highlight_override.rb
jpslav: good idea.

spec/requests/api/v0/highlights_controller_spec.rb
jpslav: I left as one test since very related but I DRYed it up with a loop.

steveburkett (approved):

PR #23, opened 2019-11-22
service limits

steveburkett:
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/686

Exception when adding a highlight thru Postman when exceeding the maximum number of highlights for a user (I set to a low number like 2, and then added another)
image

Few images showing counts incremented in postgres
image

image

Code coverage of close to 100% for all of service limits:
image


app/services/service_limits.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def self.get_check(highlights)
+    ActiveRecord::Base.transaction do
+      validate_max_gets(request_highlights_count: highlights.count)
+
+      yield highlights
+    end
+  end
+
+  def self.create_check(user_uuid, inbound_binding)
```
jpslav: think this shouldn't be based on the binding since the binding is versioned but this class is not.  passing in the `annotation:` might be better

app/services/service_limits.rb
```diff
@@ -0,0 +1,100 @@
...
+    ActiveRecord::Base.transaction do
+      user = User.find_or_create_by(id: user_uuid)
+
+      validate_max_highlights(user: user, source_id: inbound_binding.source_id)
+      validate_note_chars(user: user, note: inbound_binding.annotation)
+
+      yield inbound_binding
```
jpslav: Why wrap the creation?  Why not just

```ruby
def self.create_check(...)
  validate_max...
  validate_note_...
end
```

The raise will prevent the model from being created, and removing the yield would remove complexity and probably make testing easier.

app/services/service_limits.rb
```diff
@@ -0,0 +1,100 @@
...
+      yield highlights
+    end
+  end
+
+  def self.create_check(user_uuid, inbound_binding)
+    ActiveRecord::Base.transaction do
+      user = User.find_or_create_by(id: user_uuid)
+
+      validate_max_highlights(user: user, source_id: inbound_binding.source_id)
+      validate_note_chars(user: user, note: inbound_binding.annotation)
```
jpslav: stick with `annotation` everywhere instead of `note`?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -9,7 +9,9 @@ def create
     inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
     render(json: error, status: error.status_code) and return if error
 
-    model = inbound_binding.create_model!(user_uuid: current_user_uuid)
+    ServiceLimits.create_check(current_user_uuid, inbound_binding) do |binding|
+      model = binding.create_model!(user_id: current_user_uuid)
+    end
```
jpslav, line 15: Until I realized that the `validate_blah` methods inside `ServiceLimits` were updating fields themselves (which need a transaction for rollback behavior), it wasn't clear to me why the `create_model!` call was being passed in in a block.  Seemed like the `ServiceLimits` call could have just been a line on its own.

So I'm wondering if some renaming/rephrasing is warranted.  E.g. it can be framed more as a guard around the create call:

```ruby
model = ServiceLimits.with_create_protection do
  inbound_binding.create_model!(user_id: current_user_uuid)
end
```

```ruby
class ServiceLimits
  def self.with_create_protection
    ActiveRecord::Base.transaction do 
      yield.tap do |model| # more fun ways of using `tap` ;-)
        track_and_validate_max_highlights(model) 
        track_and_validate_max_annotation_chars(model)
      end
    end
  end
end
```

By creating the model right away with the `yield` call, you don't need to pass args in to `with_create_protection` that you are also using when you create the model, you can just pull the data from the created model. This lets you not repeat `user = User.find_or_create_by(id: user_uuid)` and do less passing around of the binding which I see as good because the binding is API-specific but the `ServiceLimits` class is not.

I also like `track_and` in the private method names because it makes it more clear that the method is updating records (more than just validating values).

Your method as currently written as well as what I just wrote above also suffer from an assumption that the line that creates the model is using `create!` instead of `create`.  If someone didn't use the exception-raising `create!` variety, both of our implementations would return unsaved models that we'd then use to update tracked service limit values.  My implementation would be improved by:

```ruby
class ServiceLimits
  def self.with_create_protection
    ActiveRecord::Base.transaction do
      model = yield # if this is using `create!` and there's an error, the exception will cause us to bail here

      # however if it was just a `create` call we need other checks
      raise ArgumentError, "Block did not yield an active record model" unless model.is_a?(Highlight)
      raise <something else> if model.errors?

      track_and_validate_max_highlights(model) 
      track_and_validate_max_annotation_chars(model)
    
      model
    end
end
```

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -19,32 +21,37 @@ def index
...
   end
 
   def update
     inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::HighlightUpdate)
     render(json: error, status: error.status_code) and return if error
 
-    model = inbound_binding.update_model!(@highlight)
+    ServiceLimits.update_check(current_user_uuid, inbound_binding) do |binding|
+      model = binding.update_model!(@highlight)
+    end
```
jpslav, line 48: Ditto above...

```ruby
model = ServiceLimits.with_update_protection { inbound_binding.update_model!(@highlight) }
```

app/models/user.rb
```diff
@@ -0,0 +1,12 @@
+class User < ApplicationRecord
+  has_many :highlights, dependent: :destroy
+  has_many :user_sources, dependent: :destroy
+
+  def increment_num_highlights(by: 1)
+    update(num_highlights: num_highlights + by)
+  end
+
+  def increment_num_annotation_characters(by: 1)
```
jpslav: no default here? (`1` is a reasonable default for `increment_num_highlights` but not as an expected annotation length)

app/models/user_source.rb
```diff
@@ -0,0 +1,9 @@
+class UserSource < ApplicationRecord
+  belongs_to :user
+
+  scope :by_source_id, ->(sourrce_id) { where(source_id: source_id) }
```
jpslav: typo

app/services/service_limits.rb
```diff
@@ -0,0 +1,100 @@
...
+  MAX_NOTE_CHARS_PER_USER = 300_000
+  MAX_CHARS_PER_NOTE = 1_000
+  MAX_HIGHLIGHTS_PER_GET = 300
+
+  class ServiceLimitsError < StandardError; end
+  class ExceededMaxHighlightsPerGet < ServiceLimitsError; end
+  class ExceededMaxHighlightsPerUser < ServiceLimitsError; end
+  class ExceededMaxHighlightsPerUserPerSource < ServiceLimitsError; end
+  class ExceededMaxNoteCharsPerUser < ServiceLimitsError; end
+  class ExceededMaxNoteChars < ServiceLimitsError; end
```
jpslav: Add custom error messages to these so that users who hit them see the reason?

db/migrate/20191120222312_create_user_and_user_sources.rb
```diff
@@ -0,0 +1,17 @@
+class CreateUserAndUserSources < ActiveRecord::Migration[5.2]
+  def change
+    create_table :users, id: :uuid do |t|
+      t.integer :num_annotation_characters
```
jpslav: `default: 0` as well?

app/services/service_limits.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      yield inbound_binding
+    end
+  end
+
+  def self.delete_reset(user_uuid, highlight)
+    ActiveRecord::Base.transaction do
+      user = User.find_or_create_by(id: user_uuid)
+
+      user.reset_max_counts(user: user, highlight: highlight)
```
jpslav: not a call on `user`

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -19,32 +21,37 @@ def index
...
 
     response_binding = Api::V0::Bindings::Highlight.create_from_model(model)
     render json: response_binding, status: :ok
   end
 
   def destroy
-    @highlight.destroy!
+    ServiceLimits.delete_reset(current_user_uuid, @highlight, &:destroy!)
```
jpslav: don't need to pass in `current_user_uuid` b/c user is available via the passed-in model.  

Do we need to pass in `destroy!` or can we just call it directly?

Maybe a more wordy method name?  `ServiceLimits.destroy_and_reset_counts(@highlight)`

jpslav (changes requested):

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -1,25 +1,28 @@
+# The primary controller for this API that implements CRUD for highlights
+#
+# See Api::V0::HighlightsSwagger for documentation
 class Api::V0::HighlightsController < Api::V0::BaseController
   before_action :render_unauthorized_if_no_current_user
 
-  before_action :get_highlight, only: [:update, :destroy]
-
-  # See Api::V0::HighlightsSwagger for documentation
+  before_action :set_highlight, only: [:update, :destroy]
```
jpslav, line 7: lol you changed it back after I changed it from `set` to `get` :-)

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -40,22 +43,31 @@ def update
     inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::HighlightUpdate)
     render(json: error, status: error.status_code) and return if error
 
-    model = inbound_binding.update_model!(@highlight)
+    updated_highlight = service_limits.with_update_protection(prev_annotation_length: @highlight.annotation&.length) do
```
jpslav: Seems like a responsibility leak to have to know that the service limits cares about highlight annotation leak.  Can we instead pass in the highlight and let `with_update_protection` decide what it needs?  If `ServiceLimits` is really only concerned with highlights I think it'd be fine to just pass in the HL.  If it is responsible for other limits, the method could be `with_highlight_update_protection(@highlight)`

app/models/user.rb
```diff
@@ -0,0 +1,12 @@
+class User < ApplicationRecord
+  has_many :highlights, dependent: :destroy
+  has_many :user_sources, dependent: :destroy
+
+  def increment_num_highlights(by: 1)
+    update(num_highlights: [num_highlights + by, 0].max)
+  end
+
+  def increment_num_annotation_characters(by: 1)
+    update(num_annotation_characters: [num_annotation_characters + by, 0].max)
```
jpslav: interested in hearing how increments into negative territory would happen; and rationale for this approach vs validating that the number stays non negative.

app/services/highlights_info.rb
```diff
@@ -30,7 +30,7 @@ def avg_highlights_per_user
           ( SELECT
               COUNT(*) AS highlights_count
             FROM highlights
```
jpslav, line 32: Is it more efficient to use the new `User.num_highlights` field rather than having PG count all highlights?

app/services/highlights_info.rb
```diff
@@ -44,7 +44,7 @@ def max_num_highlights_any_user
...
```
jpslav, line 46: Ditto, use `User.num_highlights` instead of counting

app/services/service_limits.rb
```diff
@@ -0,0 +1,124 @@
...
+  MAX_HIGHLIGHTS_PER_SOURCE_PER_USER = 300
+  MAX_HIGHLIGHTS_PER_USER            = 20_000
+  MAX_ANNOTATION_CHARS_PER_USER      = 300_000
+  MAX_CHARS_PER_ANNOTATION           = 1_000
+
+  class ServiceLimitsError < StandardError; end
+
+  class ExceededMaxHighlightsPerUser < ServiceLimitsError
+    def message
+      "Exceeded (#{MAX_HIGHLIGHTS_PER_USER} max highlights per user"
```
jpslav: All of these messages have an unclosed open parenthesis.  Might also read better if the limit is at the end of the sentences.

app/services/service_limits.rb
```diff
@@ -0,0 +1,124 @@
...
+    ActiveRecord::Base.transaction do
+      yield.tap do |model|
+        reset_max_counts_for_deleted_highlight(model)
+      end
+    end
+  end
+
+  private
+
+  def build_user
```
jpslav: "build" sounds a little like "create" when really it is "find" or "get".  Could also just say `def user`

app/services/service_limits.rb
```diff
@@ -0,0 +1,124 @@
...
+        track_and_validate_annotation_chars(model) if model.annotation.present?
+      end
+    end
+  end
+
+  def with_update_protection(prev_annotation_length:)
+    ActiveRecord::Base.transaction do
+      yield.tap do |model|
+        raise ActiveRecord::RecordInvalid, 'Block did not yield an active record model' unless model.is_a?(Highlight)
+        raise ActiveRecord::RecordInvalid, 'Model errors detected' if model.errors.present?
```
jpslav: Would be good to put the errors in the raised message (like the bang crud methods do).  Does this do it? https://gist.github.com/benshimmin/9401836#file-gistfile1-rb-L9

app/services/service_limits.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        reset_max_for_annotations(user: build_user, by: -prev_annotation_length) if prev_annotation_length.present?
```
jpslav: A "max" cannot be reset.  It is always the max, the constant defined.  Also I don't think this is really resetting per se?  

Can this all be done in `track_and_validate_annotation_chars` using the model's dirty state? (e.g. `annotation_length_was`)

app/services/service_limits.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    if user.num_annotation_characters + note_length < MAX_ANNOTATION_CHARS_PER_USER
+      reset_max_for_annotations(user: user, by: note_length)
+    else
+      raise ExceededMaxAnnotationCharsPerUser
+    end
+  end
+
+  def reset_max_counts_for_deleted_highlight(highlight)
+    user = build_user
```
jpslav: ditto elsewhere if `build_user` is just `user` we can get rid of this line

app/services/service_limits.rb
```diff
@@ -0,0 +1,124 @@
...
+  end
+
+  private
+
+  def build_user
+    @user ||= User.find_or_create_by(id: @user_id)
+  end
+
+  def track_and_validate_max_highlights(new_highlight)
+    user = build_user
```
jpslav: can get rid of this line if `def build_user` becomes `def user`

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -40,22 +43,31 @@ def update
...
+    end
 
-    response_binding = Api::V0::Bindings::Highlight.create_from_model(model)
+    response_binding = Api::V0::Bindings::Highlight.create_from_model(updated_highlight)
     render json: response_binding, status: :ok
   end
 
   def destroy
-    @highlight.destroy!
+    service_limits.with_delete_reincrement do
```
jpslav: is something being "incremented again" when deleting?  Seems like not.  In which case this method name seems not quite right.  `service_limits.track_deletion`?  or `with_delete_tracking` if you want to use the same pattern

app/services/service_limits.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        reset_max_for_annotations(user: build_user, by: -prev_annotation_length) if prev_annotation_length.present?
+        track_and_validate_annotation_chars(model) if model.annotation.present?
+      end
+    end
+  end
+
+  def with_delete_reincrement
+    ActiveRecord::Base.transaction do
+      yield.tap do |model|
```
jpslav, line 65: Need check that model is a highlight?  like have above.  Need check that deletion actually happened? (no errors, etc)

app/services/service_limits.rb
```diff
@@ -0,0 +1,124 @@
...
+    end
+
+    if user.num_annotation_characters + note_length < MAX_ANNOTATION_CHARS_PER_USER
+      reset_max_for_annotations(user: user, by: note_length)
+    else
+      raise ExceededMaxAnnotationCharsPerUser
+    end
+  end
+
+  def reset_max_counts_for_deleted_highlight(highlight)
```
jpslav: ditto thoughts about "can't reset a max count".  `track_counts_for_deleted_highlight`?

db/migrate/20191120222312_create_user_and_user_sources.rb
```diff
@@ -0,0 +1,16 @@
+class CreateUserAndUserSources < ActiveRecord::Migration[5.2]
+  def change
+    create_table :users, id: :uuid do |t|
+      t.integer :num_annotation_characters, default: 0
+      t.integer :num_highlights, default: 0
+    end
+
+    create_table :user_sources, id: :uuid do |t|
+      t.references :user, type: :uuid, index: true, foreign_key: true
+      t.string :source_id, null: false
```
jpslav, line 10: I just realized this table needs `source_type` too.  An `openstax_page` `UserSource` with `source_id` `foo` is distinct from a record with `foo` but a different `source_type`.

Also this table should have a DB uniqueness constraint on `user_id`, `source_id`, and `source_type`.

lib/tasks/clear_service_limits.rake
```diff
@@ -0,0 +1,10 @@
+desc <<-DESC.strip_heredoc
+  Clear out all service limits, resetting to 0
+DESC
+task clear_service_limits: :environment do
+  user_updated_count = User.update_all(num_annotation_characters: 0, num_highlights: 0)
+  user_source_updated_count = UserSource.update_all(num_highlights: 0)
+
+  puts "Reset #{user_updated_count} users records to 0."
+  puts "Reset #{user_source_updated_count} user_sources records to 0."
+end
```
jpslav: I don't know if we need this one.  Maybe make a card for later for a rake task that recalculates counts (in case something gets messed up)?  I think if we wanted to increase the limits for a certain user we'd just reset for all users.

spec/rails_helper.rb
```diff
@@ -1,5 +1,9 @@
 require 'simplecov'
-SimpleCov.start 'rails'
+SimpleCov.start 'rails' do
+  add_filter %r{^/app\/bindings/}
+  add_filter { |src| src.filename =~ /swagger/ }
+  add_filter { |src| src.filename =~ /scout/ }
+end
```
jpslav, line 6: are we reporting coverage? can we?

jpslav (changes requested):

app/controllers/api/v0/highlights_controller.rb
steveburkett: remove the prev length...because rails does indeed have a prior to save method for an attribute!

app/services/highlights_info.rb
steveburkett: thought of that too...but also thought this could be more official...here...we're really counting the highlights.  postgres will do this internally, i would expect it to be efficient.   also..perhaps...in some situation we decide to reset the num_highlights/note chars...this would remain as the authoritative count.

app/models/user.rb
steveburkett: it does make sense to fail early if this happens, perhaps thru a validation.  i just figured that the room for error here was higher and we'd not worry too much about it.  If you feel that we need to tighten it up..perhaps you're right and we should have a validation here.  what would you like to do?

app/services/highlights_info.rb
steveburkett: same as above ^

app/services/highlights_info.rb
steveburkett: will performance test this as part of https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/625

app/models/user.rb
steveburkett: reimplemented as a rails.logger.error before_validation.

spec/rails_helper.rb
steveburkett: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/711

lib/tasks/clear_service_limits.rake
steveburkett: deleted

app/controllers/api/v0/highlights_controller.rb
steveburkett: the use for "reincrement" was because the destroy/delete action "re" increments the server limits.

app/services/service_limits.rb
```diff
@@ -0,0 +1,125 @@
...
+  end
+
+  def with_update_protection
+    ActiveRecord::Base.transaction do
+      yield.tap do |model|
+        raise ArgumentError, 'Block did not yield an active record model' unless model.is_a?(Highlight)
+        raise ActiveRecord::RecordInvalid.new(model) if model.invalid?
+
+        prev_annotation_length = model.annotation_before_last_save.length
+        reset_max_for_annotations(user: user, by: -prev_annotation_length) if prev_annotation_length.present?
```
jpslav: Can we move these two lines into `track_and_validate_annotation_chars`?  I think we should be able to.

app/services/service_limits.rb
```diff
@@ -0,0 +1,125 @@
...
+    user.increment_num_highlights(by: -1)
+    reset_max_for_annotations(user: user, by: -highlight.annotation.length)
+
+    user_source = UserSource.find_by(user_id: highlight.user.id, source_id: highlight.source_id)
+    user_source.increment_num_highlights(by: -1)
+  end
+
+  def reset_max_for_annotations(user:, by:)
+    user.increment_num_annotation_characters(by: by)
+  end
```
jpslav: I don't think the `reset_max_for_annotations` private method is getting us much.  Can we instead just call `user.increment_num_annotation_characters(by: by)` in its place?  Will make it fall in line more with our other direct calls to `user_source.increment_num_highlights(by: 1)` and similar.  I also think the `increment_num_annotation_characters` method name is clearer than `reset_max_for_annotations`.

jpslav (changes requested):

jpslav (approved):

PR #24, opened 2019-11-22
Fix swagger

jpslav:
There were some issues with swagger that we noticed when trying to use it to generate clients and use it in the Swagger UI.  

* Some parameters were marked as required and read only which isn't allowed
* There were some half-done change to rename `UpdateHighlight` to `HighlightUpdate` 
* The way the `GetHighlights` was defined in a swagger_model and embedded as a query parameter didn't sit well with the UI or client generators.  For clients, we need that API to be well-defined in the `swagger_path` declaration; for developers, we want a `swagger_model` so that we can generate bindings that do validation on the query parameters.  I of course didn't want to define the parameters twice so I wrote a [wrapper method](https://github.com/openstax/swagger-rails/blob/6965c476297a2a4604ddd2e600b76c4721318d5c/lib/openstax/swagger/swagger_blocks_extensions.rb#L16-L40) in `openstax_swagger` that calls `swagger_path` and then turns around and automatically calls `swagger_model` with the parameters from the `swagger_path` call. This lets us define the "GET highlights" documentation with a `swagger_path` call and gets us a `GetHighlights` binding automatically.
* At one point there became a circular dependency thing with the controller file using bindings generated from the controller file's documentation.  So I moved the controller swagger documentation into its own file as a peer to the controller file.  (so now in addition to `highlights_controller.rb` we have `highlights_swagger.rb`) . Then I did the same for the Info controller.  Then I thought hey why do we have the models in a separate file and so I moved them into these dedicated swagger files.  I continued the cleanup as you'll see in the PR.


steveburkett:
nice cleanup

steveburkett (approved):

PR #28, opened 2019-12-01
Summary endpoint

jpslav:
Summary endpoint


spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -439,10 +439,53 @@
...
+    end
+
+    context 'when a user is logged in' do
+      before { stub_current_user_uuid(user_uuid) }
+
+      it 'returns good counts' do
+        get summary_path, params: query_params
+        expect(response).to have_http_status(:ok)
+        expect(json_response[:counts_per_source]).to eq({
+          "#{source_id}".to_sym => 3,
```
steveburkett: am i seeing this wrong?  looks like there should be 4 HLs for source_id and user_uuid, one for each of 2 & 3.

steveburkett: should this show some examples of filtering and HL counts, across a few users, source_id, source_type, and color variations?

config/initializers/highlight_override.rb
```diff
@@ -117,3 +117,23 @@ def update_model!(model)
     model.tap(&:save!)
   end
 end
+
+Api::V0::Bindings::GetHighlightsSummaryParameters.class_exec do
+  def summarize(user_uuid:)
+    highlights = ::Highlight.by_user(user_uuid)
```
steveburkett, line 123: kinda feels like this class (or a service class) should do this stuff, and have unit tests wrapped around it for all the filtering variations and grouping.  and then the controller can be kept golden path.

steveburkett (approved):
looks good, just a few minor questions/comments.

steveburkett:
would be cool to see a screenshot of hitting the endpoint and json results in the PR, as well as a link to the story card.

config/initializers/highlight_override.rb
jpslav: We don’t have that for other examples in this file do we?

spec/requests/api/v0/highlights_controller_spec.rb
jpslav: If we change highlight 5 to have a different user then we’d be testing filtering by user better.  Already tests color variation (there’s the red one in there not counted).  We could add a HL with a different source_type.  There’s no filtering by source_id on the summary endpoint

jpslav: added testing of filtering by user, can't test filtering by source_type b/c there is only one allowed source_type

PR #34, opened 2019-12-09
Scope neighbor checks to user

jpslav:
Fixes bug where neighbor highlight checks weren't scoped to user

### DIFF
--- a/app/models/highlight.rb
+++ b/app/models/highlight.rb
@@ -48,8 +48,8 @@ class Highlight < ApplicationRecord
 
   protected
 
-  def all_from_scope_and_source
-    Highlight.where(scope_id: scope_id).where(source_id: source_id)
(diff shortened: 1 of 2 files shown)


app/models/highlight.rb
```diff
@@ -48,8 +48,8 @@ def normalize_ids
 
   protected
 
-  def all_from_scope_and_source
-    Highlight.where(scope_id: scope_id).where(source_id: source_id)
+  def all_mine_from_scope_and_source
+    Highlight.where(user_id: user_id).where(scope_id: scope_id).where(source_id: source_id)
```
steveburkett, line 52: this could be Highlight.by_user(user_id).where.... (using a by_user scope we already have)

steveburkett (approved):

PR #35, opened 2019-12-11
Allow multiple colors on HL get and summary

jpslav:
Allow multiple colors on HL get and summary

### DIFF
--- a/app/controllers/api/v0/highlights_swagger.rb
+++ b/app/controllers/api/v0/highlights_swagger.rb
@@ -278,12 +278,16 @@ class Api::V0::HighlightsSwagger
         end
       end
       parameter do
-        key :name, :color
+        key :name, :colors
         key :in, :query
-        key :type, :string
+        key :type, :array
+        key :collectionFormat, :csv
         key :required, false
-        key :enum, VALID_HIGHLIGHT_COLORS
-        key :description, 'Limits results to this highlight color.'
+        key :description, 'Limits results to these highlight colors.'
+        items do
+          key :type, :string
+          key :enum, VALID_HIGHLIGHT_COLORS
+        end
       end
       parameter do
         key :name, :page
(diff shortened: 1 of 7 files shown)


app/models/highlight.rb
```diff
@@ -28,6 +28,7 @@ class Highlight < ApplicationRecord
 
   scope :by_source_type, ->(source_type) { where(source_type: source_type) }
   scope :by_color, ->(color) { where(color: color) }
+  scope :by_colors, ->(colors) { where(color: colors) }
```
steveburkett, line 31: do we still need the by_color scope?

app/controllers/api/v0/highlights_swagger.rb
```diff
@@ -278,12 +278,16 @@ class Api::V0::HighlightsSwagger
...
-        key :type, :string
+        key :type, :array
+        key :collectionFormat, :csv
         key :required, false
-        key :enum, VALID_HIGHLIGHT_COLORS
-        key :description, 'Limits results to this highlight color.'
+        key :description, 'Limits results to these highlight colors.'
+        items do
+          key :type, :string
+          key :enum, VALID_HIGHLIGHT_COLORS
```
jpslav, line 289: our version of swagger codegen does not generate validations for enums in an array, which is why I had to write them separately

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -101,6 +101,17 @@
           end
         end
 
+        context 'colors passed as comma-separated values in the query parameters' do
+          it 'handles them fine' do
+            get "/api/v0/highlights?source_type=openstax_page&scope_id=#{scope_1_id}&" \
+                "colors=yellow,pink"
```
steveburkett, line 107: does this need a test that tests that this test is not a false positive?  (e.g., testing for just yellow returns a subset of the yellow, pink?

spec/requests/api/v0/highlights_controller_spec.rb
```diff
@@ -480,6 +499,27 @@
           highlight2.source_id.to_sym => 1,
         })
       end
+
+      context 'colors passed as comma-separated values in the query parameters' do
+        it 'handles them fine' do
+          get "/api/v0/highlights/summary?source_type=openstax_page&scope_id=#{scope_1_id}&" \
+              "colors=yellow,pink"
```
steveburkett, line 506: we note that the arrays are passed in as csv, and not as standard URL array syntax somewhere right?

spec/requests/api/v0/highlights_controller_spec.rb
jpslav: This test already excludes the new `"green"` highlight

app/models/highlight.rb
jpslav: Maybe not but I think we could leave it.

spec/requests/api/v0/highlights_controller_spec.rb
jpslav: yeah, it is `key :collectionFormat, :csv` in the swagger

steveburkett (approved):

PR #39, opened 2019-12-18
simplify routing errors in logs

steveburkett:
404s/routing errors unnecessarily clog up the log files. 

This handles routing errors w/ a concise one liner (the normal Get log w/ status) and returns "bad request"

image

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/742

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -21,4 +21,6 @@ Rails.application.routes.draw do
       end
     end
   end
+
+  match '*path', via: :all, to: 'api/v0/base#error_404'
 end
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -21,4 +21,6 @@
       end
     end
   end
+
+  match '*path', via: :all, to: 'api/v0/base#error_404'
```
jpslav: seems a little weird for a catchall route to map to "api v0"?

steveburkett: truth.  putting it back the way i had it originally, in application controller. thanks!

PR #45, opened 2020-01-16
Add advisory locks to highlights

steveburkett:
This resolves the Highlights deadlock and foreign key problems by adding an advisory lock mutex around the destroy action.  This allows the destroy to complete the action of repairing the prev and next ids before being deleted **in the event** that there is a concurrent destructive action on the prev and next ids.   

associated PR https://github.com/openstax/perf-testing/pull/4

image

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/728

### DIFF
--- a/app/controllers/api/v0/highlights_controller.rb
+++ b/app/controllers/api/v0/highlights_controller.rb
@@ -10,14 +10,13 @@ class Api::V0::HighlightsController < Api::V0::BaseController
     inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
     render(json: error, status: error.status_code) and return if error
 
-    source_id = params[:highlight][:source_id]
-    Highlight.with_advisory_lock("#{current_user_uuid}#{source_id}") do
-      created_highlight = service_limits.with_create_protection do |user|
+    created_highlight = with_advisory_lock(inbound_binding) do
+      service_limits.with_create_protection do |user|
         inbound_binding.create_model!(user_id: user.id)
       end
-      response_binding = Api::V0::Bindings::Highlight.create_from_model(created_highlight)
-      render json: response_binding, status: :created
     end
+    response_binding = Api::V0::Bindings::Highlight.create_from_model(created_highlight)
+    render json: response_binding, status: :created
   end
 
   def index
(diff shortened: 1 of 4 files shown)


app/controllers/api/v0/highlights_controller.rb
```diff
@@ -10,12 +10,14 @@ def create
...
 
-    created_highlight = service_limits.with_create_protection do |user|
-      inbound_binding.create_model!(user_id: user.id)
+    source_id = params[:highlight][:source_id]
+    Highlight.with_advisory_lock("#{current_user_uuid}#{source_id}") do
+      created_highlight = service_limits.with_create_protection do |user|
+        inbound_binding.create_model!(user_id: user.id)
+      end
+      response_binding = Api::V0::Bindings::Highlight.create_from_model(created_highlight)
+      render json: response_binding, status: :created
```
jpslav: Can we pull the response stuff out of the advisory lock so that the lock is released sooner?

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -43,20 +45,22 @@ def update
     inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::HighlightUpdate)
     render(json: error, status: error.status_code) and return if error
 
-    updated_highlight = service_limits.with_update_protection do
-      inbound_binding.update_model!(@highlight)
+    highlight = Highlight.find(params[:id]) #refetch the highlight inside of the advisory lock
+    service_limits.with_update_protection do
+      inbound_binding.update_model!(highlight)
```
jpslav: Does this `update` method still need these changes? thought you said it didn't

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -10,12 +10,14 @@ def create
     inbound_binding, error = bind(params.require(:highlight), Api::V0::Bindings::NewHighlight)
     render(json: error, status: error.status_code) and return if error
 
-    created_highlight = service_limits.with_create_protection do |user|
-      inbound_binding.create_model!(user_id: user.id)
+    source_id = params[:highlight][:source_id]
+    Highlight.with_advisory_lock("#{current_user_uuid}#{source_id}") do
```
jpslav: `source_id` may not be downcased here which could lead to incorrect advisory lock names (since the downcasing happens in the model).

Can you add a private method to clean up the advisory lock call here and below?

```ruby
def with_advisory_lock(something_with_source_id)
  Highlight.with_advisory_lock("#{current_user_uuid}#{something_with_source_id.source_id}") do
    yield
  end
end
```

and then this becomes

```ruby
with_advisory_lock(inbound_binding) do
  ...
end
```

And down in `destroy` you'll have

```ruby
with_advisory_lock(@highlight) do
  ...
end
```

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -43,20 +45,22 @@ def update
...
-
     response_binding = Api::V0::Bindings::Highlight.create_from_model(updated_highlight)
     render json: response_binding, status: :ok
   end
 
   def destroy
-    service_limits.with_delete_tracking do
-      @highlight.destroy!
+    Highlight.with_advisory_lock("#{current_user_uuid}#{@highlight.source_id}") do
+      @highlight.reload
```
jpslav: A short comment may help future coders understand why reloading is necessary

app/controllers/api/v0/highlights_controller.rb
```diff
@@ -43,20 +45,22 @@ def update
...
+      service_limits.with_delete_tracking do
+        @highlight.destroy!
+      end
+      head :ok
```
jpslav: ditto on pulling out response and on using private `with_advisory_lock`

jpslav (changes requested):

app/controllers/api/v0/highlights_controller.rb
jpslav: also this private method can duplicate the `source_id` normalization present in the model to make sure it has been applied.

jpslav (approved):

PR #48, opened 2020-02-10
increase service limits

steveburkett:
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/826

PR #49, opened 2020-02-19
Missed swagger generated binding file for increased page limit

steveburkett:
This file was missed, it is a swagger generate file, part of this story, this previous PR: 

https://github.com/openstax/highlights-api/pull/48

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/826

(we might close this and merge this change in w/ the /info stuff instead)


TomWoodward (changes requested):
@steveburkett there is no reason that the per_page has to match the service limit, we can close this PR

PR #50, opened 2020-02-19
more info

steveburkett:
Add more data points to the /info endpoint so that product can get improved insights into highlights usage. 

This also includes a left out file from the service limits swagger generation for per_page.

https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/832

image


TomWoodward (changes requested):
the info stuff all looks fine to me, the changes to app/bindings/api/v0/bindings/get_highlights_parameters.rb aren't necessary though.

TomWoodward (approved):
ideally we'd get @jpslav  to read this too, but lets merge before friday regardless

PR #61, opened 2020-05-06
Curated highlights

TomWoodward:


lib/tasks/curator.rake
```diff
@@ -0,0 +1,21 @@
+namespace :curator do
+  desc <<-DESC.strip_heredoc
+    Sets the curator for a given scope, curated highlights are available by querying the set
+    "curated:openstax".
+
+    eg: rake curator:set[scope_uuid,curator_uuid]
+  DESC
+  task :set, [:scope_id, :curator_id] => :environment do |task, args|
+    CuratorScope.where(:scope_id => args[:scope_id]).delete_all
```
jpslav: 1. More modern to use more compact syntax: `where(scope_id: args[:scope_id])`
2. Use `destroy_all` instead of `delete_all`.  The `destroy` methods call callbacks on the models (e.g. there could be a `before_destroy` callback that checks that it is ok to delete the model).  The `delete` methods don't hit these callbacks.

lib/tasks/curator.rake
```diff
@@ -0,0 +1,21 @@
...
+    CuratorScope.create(args.to_hash)
+  end
+
+  desc <<-DESC.strip_heredoc
+    Clears the curator for a given scope
+
+    eg: rake curator:clear[scope_uuid]
+  DESC
+  task :clear, [:scope_id] => :environment do |task, args|
+    CuratorScope.where(:scope_id => args[:scope_id]).delete_all
```
jpslav: Ditto on `scope_id: args[:scope_id]` and `destroy_all`

Do we want only one curator per scope?  If so we should have some database constraints and constraints on the model.  I can point you in the right direction if this is the case.

lib/v0_bindings_extensions.rb
```diff
@@ -83,11 +83,21 @@ def self.create_from_query_result(query_result)
 
   Api::V0::Bindings::GetHighlightsParameters.class_exec do
     def query(user_id:)
-      highlights = ::Highlight.by_user(user_id)
+      users = []
+
+      if sets.blank? or sets.include?('user:me')
+        users.push(user_id)
+      end
+      if  sets.to_a.include?('curated:openstax') and scope_id.present?
```
jpslav: 1. Extra space after `if`
2. I think `sets` is already an array, so I don't think you need `.to_a`

lib/v0_bindings_extensions.rb
```diff
@@ -83,11 +83,21 @@ def self.create_from_query_result(query_result)
...
+        curator_scope = CuratorScope.find_by(:scope_id => scope_id)
```
jpslav: `scope_id: scope_id`

lib/v0_bindings_extensions.rb
```diff
@@ -83,11 +83,21 @@ def self.create_from_query_result(query_result)
 
   Api::V0::Bindings::GetHighlightsParameters.class_exec do
     def query(user_id:)
-      highlights = ::Highlight.by_user(user_id)
+      users = []
```
jpslav: `users` name implies this is an array of `User`.  Recommend `user_ids = []` instead.

lib/v0_bindings_extensions.rb
```diff
@@ -83,11 +83,21 @@ def self.create_from_query_result(query_result)
...
+
+      if sets.blank? or sets.include?('user:me')
+        users.push(user_id)
+      end
+      if  sets.to_a.include?('curated:openstax') and scope_id.present?
+        curator_scope = CuratorScope.find_by(:scope_id => scope_id)
+        users.push(curator_scope.curator_id) unless curator_scope.nil?
```
jpslav: Should there be an error if the requested curator doesn't exist?

lib/v0_bindings_extensions.rb
```diff
@@ -123,11 +133,21 @@ def update_model!(model)
...
-      highlights = ::Highlight.by_user(user_uuid)
+      users = []
+
+      if sets.blank? or sets.include?('user:me')
+        users.push(user_uuid)
+      end
+      if  sets.to_a.include?('curated:openstax') and scope_id.present?
+        curator_scope = CuratorScope.find_by(:scope_id => scope_id)
+        users.push(curator_scope.curator_id) unless curator_scope.nil?
+      end
```
jpslav: To DRY this up, recommend a `sets_to_user_ids` method somewhere or similar.

lib/v0_bindings_extensions.rb
```diff
@@ -123,11 +133,21 @@ def update_model!(model)
...
+
+      highlights = ::Highlight.by_user(users)
```
jpslav: Instead of a separate line here that kind of duplicates line 151 below, you could instead do

```ruby
filters = to_hash.except(:sets)
filters[:user] = sets_to_user_ids(sets) # or something
```

And then this line goes away and line 150 becomes

```ruby
filter.each do |key, value|
```

lib/v0_bindings_extensions.rb
```diff
@@ -153,9 +173,17 @@ def valid_colors?
         invalid_colors.empty?
       end
 
+      def invalid_sets
+        sets.present? ? sets - Api::V0::HighlightsSwagger::VALID_SETS : []
```
jpslav, line 174: If `sets` is always an array, which I think is true, this can simplify to 

```ruby
sets - Api::V0::HighlightsSwagger::VALID_SETS
```

because

```
irb(main):001:0> [] - [1,2,3]
=> []
```

jpslav (changes requested):

lib/v0_bindings_extensions.rb
TomWoodward: the to_a is a null check, i can change it to be more explicit

lib/v0_bindings_extensions.rb
TomWoodward: i wasn't sure where to put this, for now i put it in the shared monkey patch, but the whole thing makes me feel dirty.

lib/v0_bindings_extensions.rb
TomWoodward: sets can be null

lib/v0_bindings_extensions.rb
TomWoodward: no, we want to return an empty set if there is no curation, as an abstraction for the client

lib/v0_bindings_extensions.rb
```diff
@@ -165,6 +130,38 @@ def summarize(user_uuid:)
     Api::V0::Bindings::GetHighlightsSummaryParameters
   ].each do |klass|
     klass.class_exec do
+
+      def get_highlights(user_id)
+        filters = to_hash.except(:page, :per_page, :sets)
+        filters[:user] = []
+
+        if (sets.blank? or sets.include?('user:me')) and user_id.present?
```
jpslav: normally better to use `||` and `&&` instead of `or` and `and` because the former have higher operator precedence than the latter.  Here it won't have an impact b/c there are no other operators between the boolean operators, but if there were the lower precedence of `or` and `and` could cause unexpected behavior.

jpslav (changes requested):

jpslav (approved):

app/bindings/api/v0/bindings/get_highlights_parameters.rb
```diff
@@ -20,6 +20,9 @@ class GetHighlightsParameters
     # Limits results to the source document container in which the highlight was made.  For openstax_page source_types, this is a versionless book UUID. If this is not specified, results across scopes will be returned, meaning the order of the results will not be meaningful.
     attr_accessor :scope_id
 
+    # One or more sets to load data from; default is \"user:me\"
+    attr_accessor :sets
```
steveburkett, line 24: i would think a more descriptive name would be better here.  "sets" seems kinda vague.  maybe "user_sets"?

app/models/curator_scope.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
```
steveburkett, line 2: I would suggest that we start putting documentation at the top of models that explains the model, uses, concepts at a higher level.  this would be super for future devs, for the code to be more self documenting. 

eg, for this.  what is a curator scope? why do we need it?  how are records added, updated, or deleted? etc etc.

db/migrate/20200505135519_curator_scope.rb
```diff
@@ -0,0 +1,9 @@
+class CuratorScope < ActiveRecord::Migration[5.2]
+  def change
+    create_table :curator_scopes, id: :uuid do |t|
+      t.uuid :curator_id, null: false, index: true
+      t.uuid :scope_id, null: false
```
steveburkett, line 5: this is fine (a separate index), but it is possible to make it more compact:
```
t.uuid :scope_id, null: false, index: {unique: true}
```

lib/tasks/curator.rake
```diff
@@ -0,0 +1,21 @@
+namespace :curator do
+  desc <<-DESC.strip_heredoc
+    Sets the curator for a given scope, curated highlights are available by querying the set
+    "curated:openstax".
+
+    eg: rake curator:set[scope_uuid,curator_uuid]
+  DESC
+  task :set, [:scope_id, :curator_id] => :environment do |task, args|
+    CuratorScope.where(scope_id: args[:scope_id]).destroy_all
+    CuratorScope.create(args.to_hash)
```
steveburkett, line 10: who runs this?  should we check the args before they're passed into destroy and create?  (e.g., could someone mess this up badly?)

lib/v0_bindings_extensions.rb
```diff
@@ -145,6 +130,38 @@ def summarize(user_uuid:)
     Api::V0::Bindings::GetHighlightsSummaryParameters
   ].each do |klass|
     klass.class_exec do
+
+      def get_highlights(user_id)
+        filters = to_hash.except(:page, :per_page, :sets)
```
steveburkett, line 135: this seems like a lot of code "parts" that maybe should be written as chainable scopes on the model (some with parameters?)

lib/v0_bindings_extensions.rb
```diff
@@ -145,6 +130,38 @@ def summarize(user_uuid:)
...
+        filters[:user] = []
+
+        if (sets.blank? || sets.include?('user:me')) && user_id.present?
```
steveburkett, line 138: this would be a bit easier to understand if the || part were a method and it read  if sets_me && user_id.present?

lib/v0_bindings_extensions.rb
```diff
@@ -145,6 +130,38 @@ def summarize(user_uuid:)
...
+          filters[:user].push(user_id)
+        end
+
+        if sets.present? && sets.include?('curated:openstax') && scope_id.present?
```
steveburkett, line 142: a 'def build_by_user_filters' would be cool (maybe even as a separate class with unit tests)..this method has a lot of big responsibilities. 

altho...all of this already inside a class_exec, so maybe not.  not ideal we have to do it this way.  e.g., stacking unit tests around this....

lib/v0_bindings_extensions.rb
```diff
@@ -145,6 +130,38 @@ def summarize(user_uuid:)
...
+          curator_scope = CuratorScope.find_by(scope_id: scope_id)
+          filters[:user].push(curator_scope.curator_id) unless curator_scope.nil?
+        end
+
+        # if there is no curator, or user is logged out, its possible to not
```
steveburkett, line 147: its should be it's :)

lib/v0_bindings_extensions.rb
steveburkett: this string "curated:openstax' should prob. be stored somewhere (on the Highlight model perhaps) as a const.  once place to store very important strings like this

PR #67, opened 2020-07-09
allow empty annotation updates

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/758

Not sure if checking `is_a?(String)` is redundant.
Also, since this would allow whitespace only notes, if to count their characters.

### DIFF
--- a/spec/services/service_limits_spec.rb
+++ b/spec/services/service_limits_spec.rb
@@ -218,7 +218,7 @@ RSpec.describe ServiceLimits, type: :service do
       it 'will update when annotation is set to empty string' do
         expect do
           service_limits.with_update_protection do
-            highlight.annotation = ""
+            highlight.annotation = ''
             highlight.tap(&:save!)
           end
         end.to_not raise_error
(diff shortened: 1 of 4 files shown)


TomWoodward (changes requested):
the schema says that `annotation` is optional, so i think your is_a checks are correct, you should add some specs though

TomWoodward (approved):

TomWoodward:
@steveburkett i tagged you on this to make sure we're doing good railsy stuff, please take a look when you have time

spec/services/service_limits_spec.rb
```diff
@@ -209,6 +209,37 @@
...
+    context 'whitespace annotations' do
+      let!(:user) { create(:new_user, num_annotation_characters: user_annotation.length ) }
+      let!(:highlight) { create(:highlight, user: user, annotation: user_annotation) }
+
+      let(:user_annotation) { 'annotation' }
+
+      it 'will update when annotation is set to empty string' do
+        expect do
+          service_limits.with_update_protection do
+            highlight.annotation = ""
```
steveburkett: very minor point (optional). i would stay consistent w/ single (or double) quotes for non-interpreted strings if possible.

spec/services/service_limits_spec.rb
```diff
@@ -209,6 +209,37 @@
...
+            highlight.tap(&:save!)
+          end
+        end.to_not raise_error
+
+        expect(user.reload.num_annotation_characters).to eq 0
+      end
+
+      it 'counts whitespace only annotations' do
+        whitespace_annotation = " \n"
```
steveburkett: is this line really necessary?  the it comment kinda makes it clear.

steveburkett:
@TomWoodward this seems fine.  we want to allow any string, even if it's empty or just whitespace or has something.

spec/services/service_limits_spec.rb
janKozie1: I've left `" \n"` since api seems to count `\n` as one character and single quotes would interpret it literally.

PR #69, opened 2020-08-06
Temporarily remove the gt_200 query

steveburkett:


app/services/highlights_info.rb
```diff
@@ -100,18 +100,20 @@ def max_num_highlights_any_user
...
+    #   FROM
+    #     ( SELECT
+    #         COUNT(*)
+    #       FROM highlights
+    #       GROUP BY user_id, source_id
+    #         HAVING COUNT(*) > #{GREATER_THAN_200}) temp_table
+    # SQL
+    #
+    # ActiveRecord::Base.connection.select_value(query)
+    -1
```
reedstrm, line 116: How does the receiving code handle a result of -1 ?

steveburkett: @reedstrm yes.  it's an int.  i was going to use "na" but that's not an int :)    -1 tells Kim that it's a nonsense response that's not implemented.

TomWoodward (approved):

PR #72, opened 2020-08-18
Allow requesting a highlight by its id

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1250


janKozie1:
I'm having trouble generating a highlight client so I can test it with rex. 
Besides that I'm not sure who should review it. I'm sorry if I should have tagged someone else.

app/controllers/api/v0/highlights_swagger.rb
```diff
@@ -431,6 +431,36 @@ class Api::V0::HighlightsSwagger
     end
   end
 
+  swagger_path '/highlights/{id}' do
+    operation :get do
+      key :summary, 'Get a highlight without an annotation.'
```
steveburkett: this wording could be confusing...it might mean:   find and get this highlight that doesnt havent an annotation..  perhaps it's more accurate to say:  get a highlight without its annotation data?  just a thought here.  the first time i read this i thought this.

steveburkett (approved):

janKozie1:
@steveburkett, I'm not sure why the build is failing, is it because the codecov 0.1.16 was yanked and it wasn't able to download cache? 
Also, who does the merging?

PR #79, opened 2022-01-11
Update highlighting sorting

jivey:
for: openstax/unified#1888

- If the incoming prev/next ids won't cause an adjacency error, use them and don't do anything else
- If there are saved highlights that **do not** have the incoming anchor, find the relative position by parsing anchors out of the page html from archive
- If there are saved highlights that **do** have the incoming anchor, find the relative position using `content_path` (a combination of indexes to the highlighted node relative to the anchor, and `start_offset`) if it's available
- If placement can't be determined, place it at the end of the highlights so an error isn't thrown


app/models/highlight.rb
```diff
@@ -47,12 +48,118 @@ def normalize_ids
...
+    anchors = all_mine_from_scope_and_source.order(:order_in_source).map(&:anchor)
+    return unless anchors.any?
+
+    # If the highlight is inside an existing anchor, find the neighbors
+    # using content_path, otherwise find it relative to all page anchors
+    if anchors.include?(anchor)
+      set_inside_anchor_neighbors
+    elsif page_anchors_comparable?
+      set_outside_anchor_neighbors
+    elsif !prev_highlight_id && !next_highlight_id
```
TomWoodward: should this be an `else`? if the prev/next were legitimate we would have short circuited on 116 right?

app/services/page_content.rb
```diff
@@ -0,0 +1,43 @@
...
+  def s3
+    @s3 ||= OpenStax::Content::S3.new
+  end
+
+  def latest_archive_version
+    @latest_archive_version ||= s3.ls.last
+  end
+
+  def archive
+    @archive ||= OpenStax::Content::Archive.new latest_archive_version
```
TomWoodward: you should use the archive that rex uses, https://openstax.org/rex/config.json

technically each book could also have an archive override https://staging.openstax.org/rex/release.json

there is a chance if you don't use the same book that your desired book/version will not be found

its still a little janky because it won't be environmentally aware, i'm not sure how to do that without adding more config to the api service. its probably fine

TomWoodward (changes requested):
some minor comments, this looks really good

app/models/highlight.rb
jivey: Yes good catch, I think this condition was from before I added the adjacency check up there and is unnecessary.

app/services/page_content.rb
jivey: Updated to use a secrets `archive_version` setting: https://github.com/openstax/highlights-api/pull/79/commits/9a3cdd9ea3ce0f928478814afc57f28283d1eb0e

I didn't know that about the book override, interesting! Now I wonder if that's fetchable... If so I could use it when initializing the content fetch.

TomWoodward: what do you mean fetchable?

app/services/page_content.rb
```diff
@@ -0,0 +1,68 @@
...
+  def latest_archive_version
+    @latest_archive_version ||= s3.ls.last
+  end
+
+  def overriden_archive_version
+    @overriden_archive_version ||= fetch_rex_books.dig(@book_uuid, 'archiveOverride').to_s.gsub('/apps/archive/', '').presence
+  end
+
+  def archive_version
+    overriden_archive_version || Rails.application.secrets.content[:archive_version] || latest_archive_version
```
TomWoodward: yea idk if i really want to be on the hook for updating this in the config every time we upgrade to a new pipeline, i know its less than ideal to do the extra queries, but the right thing is to load it from the config.json

alternatively you could switch it over to using the archive_prefix from the `source_metadata` and that part will just do nothing for now until we complete the frontend change, that might be overall less work than messing around with all these configs

TomWoodward (changes requested):

config/secrets.yml.example
```diff
@@ -30,6 +41,16 @@ test:
...
+      archive_path: apps/archive
+      bucket_name: <%= ENV['OPENSTAX_CONTENT_BUCKET_NAME'] || 'not-a-real-bucket' %>
+      domain: openstax.org
+      s3_region: <%= ENV['OPENSTAX_CONTENT_S3_REGION'] || 'us-east-1' %>
+      s3_access_key_id: <%= ENV['OPENSTAX_CONTENT_S3_ACCESS_KEY_ID'] || 'not-a-real-key' %>
+      s3_secret_access_key: <%= ENV['OPENSTAX_CONTENT_S3_SECRET_ACCESS_KEY'] || 'not-a-real-key' %>
+      rex_release_url: <%= ENV['REX_RELEASE_URL'] || 'https://dev.openstax.org/rex/release.json' %>
+      rex_config_url: <%= ENV['REX_CONFIG_URL'] || 'https://dev.openstax.org/rex/config.json' %>
 
 production:
```
TomWoodward, line 41: do we need to add the content config block to production as well?

TomWoodward (changes requested):

config/secrets.yml.example
jivey: I'm not sure if the deploy for this just copies the example to make the secrets.yml or if it's manual and the config sticks around, either way it's probably helpful to add it to that block, I'll update it

config/secrets.yml.example
```diff
@@ -50,3 +71,13 @@ production:
       signature_public_key: |
         ssm(sso_signature_public_key)
       encryption_private_key: ssm(sso_encryption_private_key)
+  content:
+      abl_url: <%= ENV['OPENSTAX_CONTENT_ABL_URL'] || 'https://raw.githubusercontent.com/openstax/content-manager-approved-books/master/approved-book-list.json' %>
```
TomWoodward: the other prod credentials are either using that `{{ }}` notation, which i think is handled [here](https://github.com/openstax/unified-deployment/blob/cb2cbddce1f99de134b2acd643e45c161d6de7ae/apps/highlights/scripts/deployment.rb#L122), or they're loading from ssm, in either case, i think these content parameters need to be updated to work correctly

for each of these params we're going to need to figure out how to populate it... and it might be easier if there were less of them... for instance i don't actually think you need all the parameters about the direct s3 access

config/secrets.yml.example
```diff
@@ -50,3 +71,13 @@ production:
...
+      archive_path: <%= ENV['OPENSTAX_CONTENT_ARCHIVE_PATH'] %>
+      bucket_name: <%= ENV['OPENSTAX_CONTENT_BUCKET_NAME'] %>
+      domain: <%= ENV['OPENSTAX_CONTENT_DOMAIN'] || 'openstax.org' %>
+      s3_region: <%= ENV['OPENSTAX_CONTENT_S3_REGION'] %>
+      s3_access_key_id: <%= ENV['OPENSTAX_CONTENT_S3_ACCESS_KEY_ID'] %>
+      s3_secret_access_key: <%= ENV['OPENSTAX_CONTENT_S3_SECRET_ACCESS_KEY'] %>
+      rex_release_url: <%= ENV['REX_RELEASE_URL'] || 'https://openstax.org/rex/release.json' %>
```
TomWoodward: highlights deployments don't... really know what environment they're going to be used in, they know which accounts env they're supposed to use, but that isn't quite the same thing. we're either going to have to add a parameter to pull all the way down through the deployment scripts or deal with it being wrong sometimes in lower environments

TomWoodward (changes requested):

config/secrets.yml.example
jivey: S3 stuff is only needed as a fallback to grab the latest version from S3 if the REX version lookup fails, which seems pretty unlikely, so I could probably remove it.

I totally overlooked the substitution and ssm stuff, that script link helps clarify things for me... seems like I need to poke around the deployment repo.

config/secrets.yml.example
```diff
@@ -30,6 +35,10 @@ test:
         hwIDAQAB
         -----END PUBLIC KEY-----
       encryption_private_key: 'RvGHVZ/kvzUAA5Z3t68+FNhuMCJxkzv+'
+  content:
+    archive_path: <%= ENV['OPENSTAX_CONTENT_ARCHIVE_PATH'] || 'apps/archive' %>
+    domain: <%= ENV['OPENSTAX_CONTENT_DOMAIN'] || 'openstax.org' %>
+  rex_host: <%= ENV['REX_HOST'] || 'https://dev.openstax.org' %>
```
TomWoodward: so in lower tier we want this to be dynamic right?

config/secrets.yml.example
```diff
@@ -14,6 +14,11 @@ development:
...
```
TomWoodward: here too right?

TomWoodward (changes requested):

config/secrets.yml.example
jivey: I didn't do this because `config.json` isn't setup for the local proxy, but after looking into it more, I see that it'd be easy to add it to the proxy and have it return `config.archive-url.json` (it only needs REACT_APP_ARCHIVE) what do you think?

jivey:
@TomWoodward updated to use a pipelineVersion: https://github.com/openstax/highlights-api/pull/79/commits/1cee5323533787aede872cdd3b8dfae46eab2f87

TomWoodward (approved):

PR #81, opened 2022-02-11
Test GH actions with jemalloc

Dantemss:
Just trying to make sure jemalloc doesn't blow anything up


Dantemss:
Won't be necessary... probably

PR #83, opened 2022-08-02
Updated openstax_auth to enable using SSO cookies in the Authorization header

Dantemss:
For: https://github.com/openstax/unified/issues/1980
Note that openstax_auth is actually https://github.com/openstax/auth-ruby, not auth-rails.


jivey (approved):

Dantemss:
I need to poke GH actions to reactivate it I guess

PR #85, opened 2025-10-21
Update to Ruby 3.1/Rails 6.1

Dantemss:
https://openstax.atlassian.net/browse/CORE-658

Ignore the migration check, it doesn't work across ruby versions.


app/services/info_store.rb
```diff
@@ -152,15 +152,15 @@ def median_note_length
...
         FROM
           (SELECT
-             pg_column_size(annotation) 
+             pg_column_size(annotation) as column_size
              FROM highlights
              WHERE annotation IS NOT NULL) as note_size
     SQL
 
-    ActiveRecord::Base.connection.select_value(query)&.gsub(/[()]/, "").to_i
+    ActiveRecord::Base.connection.select_value(query).to_i
```
jivey, line 163: It looks like the old code expected this to potentially be nil, just double checking this is a safe cast.

app/services/info_store.rb
```diff
@@ -152,15 +152,15 @@ def median_note_length
     query = <<-SQL
       SELECT
         percentile_disc(0.5)
-        WITHIN group (order by note_size)
+        WITHIN group (order by note_size.column_size)
         FROM
           (SELECT
-             pg_column_size(annotation) 
+             pg_column_size(annotation) as column_size
```
jivey, line 158: Guessing this (good) change is from stricter parsing in the upgraded postgres version

jivey (approved):
Will need testing but looks good to me, left one small concern about the `to_i` cast.

app/services/info_store.rb
Dantemss: Well, `nil&.blah.to_i` is the same as `nil.to_i` which is`0`, so the behavior should be the same...

jivey: Forgot `to_i` is on NilClass, the JS is slowly replacing it all :P

PR #87, opened 2025-11-05
Zeitwerk compatibility

Dantemss:
https://openstax.atlassian.net/browse/CORE-658

- Moved lib/errors to individual files in app/errors
- Moved lib/v0_bindings_extensions.rb to initializers
- Removed manual lib requires, since highlights adds lib to the eager_load_paths


TomWoodward (approved):
Image

